PR #20, opened 2015-01-06
convert the encoding to UTF-8 before writing out the content

yihui:
Without the conversion, multi-byte characters do not work in HTML widgets. I have tested it with the DT package.

This fixes the third issue reported at rstudio/shiny-examples#2.

### DIFF
--- a/R/html_print.R
+++ b/R/html_print.R
@@ -96,6 +96,7 @@ save_html <- function(html, file, background = "white", libdir = "lib") {
             rendered$html,
             "</body>",
             "</html>")
+  html <- enc2utf8(html)
 
   # write it
   writeLines(html, file, useBytes = TRUE)


jjallaire:
Looks good to me. @jcheng5  ?

We should also accompany this with a version bump so that htmlwigets can require the updated version explicitly.

jcheng5:
@yihui and I discussed my concerns (similar to @hadley's) at length over IM. We were seeing some really strange behavior from R, such that it seems apparent that neither he nor I actually understand precisely how text encoding in R works.

See http://rpubs.com/jcheng/encoding-weirdness, basically the one case where you'd expect enc2utf8 to be most useful (giving it an explicit latin1 string) doesn't seem to work. This patch seems to work OK because if the encoding is "unknown" then enc2utf8 seems to guess heuristically (I haven't looked at the source code yet, but that's empirically what I'm seeing).

jcheng5:
Forget my previous rpubs, that example breaks because the character I chose is in Windows-1252 but not in Latin1.

jcheng5:
After talking in depth with @hadley and a bunch more investigation on my own, I've realized some things that may help.

As part of the tag rendering process, the tag output is accumulated in an anonymous file connection (see [here](https://github.com/rstudio/htmltools/blob/60f5259d8fea20f9fa13d3e7612db2ab174cce75/R/tags.R#L449-L454)). The act of stuffing the data into a file connection and then taking it back out, causes the input strings to be translated to native encoding.

That's good in the sense that we get predictable behavior regardless of the input encoding; everything is coerced to native encoding. However, that means it's totally impossible to render any tags that contain characters that can't be represented in native encoding. For example, `tags$div("丑")` (or even the equivalent `tags$div("\u4E11")`) fails on Windows with Windows-1252.

This alternative change will coerce everything to UTF-8 instead.

screen shot 2015-01-08 at 12 59 21 pm

Even mixing UTF-8 and system encoded data works fine:
screen shot 2015-01-08 at 1 00 41 pm

yihui:
Actually I have a question from day one: why did you choose to render tags through a file connection? Reading/writing characters from/to files can be lossy for those characters that are not expressible in native encoding, unless you read/write raw bytes. Is it difficult to avoid file i/o and just render the tags in memory? If that is possible, I guess my one-line patch here should just work, and we will not need the extra layers of `enc2utf8()` + `readLines(, encoding = 'UTF-8')` for every piece of the tags.

Personally I'm not terribly worried about the characters that cannot be represented in native encoding. I believe this should be rare from the user's perspective, because I guess most users just work in their own language.

jcheng5:
We needed to use a connection in order for it not to be unusably slow. I tried textConnection first but it turns out that internally it just uses a character vector, so it's also unusably slow. The docs for textConnection say to use an anonymous file() connection if you have large amounts to write.

It just seems bizarre to me that on an en-US install of Windows, you can do `"䀑"` just fine but `div("䀑")` is corrupt. Sure, an en-US user won't be typing that into their source code, but their data can come from anywhere.

What do we lose with my fix? We're still doing enc2utf8 on the same number of bytes (albeit split into many more calls) and if it turns out it's really slow we can easily optimize it. The main downside you pointed out yesterday is `as.character(tags)` will now return properly tagged UTF-8 encoded strings instead of "unknown"-tagged native encoded strings, which may interfere with grep/gsub.

jcheng5:
I may have misunderstood something from your last comment, can you clarify something for me?

> Is it difficult to avoid file i/o and just render the tags in memory? If that is possible, I guess my one-line patch here should just work

If the tags are just rendered in memory, and the tags originally had mixed UTF-8 and native-encoded input, what do you envision the one-line patch receiving as its input value?

yihui:
It is fine for enc2utf8() to take input with mixed encodings, e.g.

```r
x = c("\u4E11", "\300", "œ", 'o')
x
Encoding(x)
y = enc2utf8(x)
y
Encoding(y)
```

```
> x = c("\u4E11", "\300", "œ", 'o')
> x
[1] "丑" "À"  "œ"  "o" 
> Encoding(x)
[1] "UTF-8"   "latin1"  "latin1"  "unknown"
> y = enc2utf8(x)
> y
[1] "丑"     "À"      "\u009c" "o"     
> Encoding(y)
[1] "UTF-8"   "UTF-8"   "UTF-8"   "unknown"
> sessionInfo()
R version 3.1.2 (2014-10-31)
Platform: x86_64-w64-mingw32/x64 (64-bit)

locale:
[1] LC_COLLATE=English_United States.1252 
[2] LC_CTYPE=English_United States.1252   
[3] LC_MONETARY=English_United States.1252
[4] LC_NUMERIC=C                          
[5] LC_TIME=English_United States.1252    

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

loaded via a namespace (and not attached):
[1] tools_3.1.2
```

I do not have strong objections to the patch of explicitly converting every piece to UTF8. Besides the possible downside you mentioned, performance may or may not be a real issue (I need to benchmark it).

What I meant by rendering tags in memory is to collect the character strings in memory instead of writing them out to a file, e.g. for a tag `x = list(name = 'a', attribs = list(href = '#'), children = 'foo')`, the rendering might be `paste0('<', x$name, paste(x$attribs, ...) '>', x$children, '</', x$name, '>')`. I imagine rendering in memory should not be slower than writing to file(). By looking at `doRenderTags()`, it is not clear to me how to rewrite it, though.

jcheng5:
I see. I was imagining that as.character always gives you a single element character vector, it seems much nicer for users so there's no ambiguity whether the different character vector elements represent distinct lines or not.

On my Core i3 Linux box:

```r
> system.time(for (i in 1:1000000) enc2utf8("hi"))
   user  system elapsed 
  0.084   0.000   0.081 
```

Collecting the character strings in memory was what we effectively did before, and it was literally thousands of times slower for large but not unreasonable documents (like some of the examples from Andrew Clark), because concatenating immutable strings involves crazy amounts of memory allocation and copying. However, that collection was happening iteratively, not recursively like you have here; maybe that would make a difference.

yihui:
Okay, I did not know it had been done in history, and writeChar() could be so much faster. I guess `enc2utf8('hi')` is not a realistic example since it is purely ASCII. I tried this instead (with my Core i7 Linux):

```r
x = c("\u4E11", "\300", "œ", 'o')
system.time(for (i in 1:1000000) enc2utf8(x))
#    user  system elapsed 
#  1.382   0.000   1.323 
```

I guess it should not be a real issue since it is unlikely that someone needs to render so many tags. Anyway, could you submit your patch as a new PR? I can test it on Rmd/htmlwidgets examples to see if it really makes a difference. Thanks!

PR #21, opened 2015-01-09
Convert all string data to UTF-8 at render time

jcheng5:
See discussion at https://github.com/rstudio/htmltools/pull/20


R/tags.R
```diff
@@ -445,13 +450,29 @@ renderTags <- function(x, singletons = character(0), indent = 0) {
...
+  # UTF-8 using enc2utf8. The rendered output will always be UTF-8
+  # encoded.
+  #
+  # We use a file() here instead of textConnection() or paste/c to
+  # avoid the overhead of copying, which is huge for moderately
+  # large numbers of calls to connWriter(). Generally when you want
+  # to incrementally build up a long string out of immutable ones,
+  # you want to use a mutable/growable string buffer of some kind;
+  # since R doesn't have something like that (that I know of),
+  # file() is the next best thing.
```
yihui: Thanks for the explanation. This is much clearer to me now.

inst/tests/test-tags.r
```diff
@@ -457,3 +457,43 @@ test_that("Non-tag objects can be coerced", {
   expect_identical(result2$singletons, "110d1f0ef6762db2c6863523a7c379a697b43ea3")
 
 })
+
+test_that("Latin1 and system encoding are converted to UTF-8", {
```
yihui: Testing these things can be very tricky, since they may be OS/locale-dependent, e.g. on Linux,

```r
> Encoding("\377")
[1] "unknown"
```

hadley: To be safe, I think it would be best to create examples with `rawToChar()` followed by an explicit setting of the encoding

jcheng5:
Encoding is hard.

jcheng5:
OK, thanks for the tips. Now passes tests on Windows (en-US), Ubuntu 14.04, Mac 10.10. @yihui do you mind testing on your Chinese Windows, please?

jcheng5:
There is one last set of issues I'm working through that cause the tests to fail on Windows with `Sys.setlocale(, "Chinese")` (for example). The existing PR works well for strings encoded using native encoding (`unknown`) and/or UTF-8. But it doesn't work well for `latin1`, unless the system encoding happens to be Latin-1 (or Windows-1252).

The reason for this is that we have a few `paste` calls in the rendering pipeline and if the inputs to paste contain Latin1 inputs but not any UTF-8, then the output is `unknown` encoded. Whereas we want Latin1 inputs to be promoted to UTF-8.

It's easy to fix this by scattering enc2utf8 calls everywhere a paste is about to be called, but that might be a little tricky to maintain (though the fact that we have tests now, helps). Another option would be to traverse the tag tree as rendering begins and call enc2utf8 everywhere then (we already traverse the tag tree here for other reasons).

R/tags.R
```diff
@@ -1,6 +1,31 @@
...
+  # http://rpubs.com/hadley/do-call2 this will inline the values, which could be
+  # potentially large strings.
+  #
+  # These lines are basically the equivalent of (for a single unnamed arg):
+  # paste(args[[1L]], sep = args[["sep"]], collapse = args[["collapse"]])
+  argNames <- ifelse(nzchar(names(args)), as.list(names(args)), 1:length(args))
+  argz <- lapply(argNames, function(x) {
+    eval(substitute(quote(args[[x]]), list(x=x)))
+  })
+  names(argz) <- names(args)
```
jcheng5: @hadley, are lines 20-24 overkill compared to just doing `do.call("paste", args)`? paste8 will be called many times over the course of a large render, but in my (non-micro-)benchmarking I wasn't able to see any difference in speed for large inputs (like `rep("a", 1e7)`).

hadley: Yes, they're total overkill.  Inlining the values is actually a slight performance _enhancement_ most of the time, because it avoids a single variable-value lookup. The downside only comes if an exception is thrown, because it takes much longer to generate the traceback.

jcheng5:
@yihui This is good to go as far as I know. Can you run one last test on it on your machine? (I've tried to duplicate your setup by setting my Windows default codepage to Chinese but RStudio still starts with en-US.1252)

yihui:
Sure. I'm be testing it in a minute.

You can put `Sys.setlocale(, 'Chinese')` in .Rprofile temporarily for testing purposes.

R/tags.R
```diff
@@ -1,6 +1,19 @@
...
+paste8 <- function(..., sep = " ", collapse = NULL) {
+  args <- c(
+    lapply(list(...), enc2utf8),
+    list(
+      sep = if (is.null(sep)) sep else enc2utf8(sep),
+      collapse = if (is.null(collapse)) collapse else enc2utf8(collapse)
+    )
+  )
+
+  do.call("paste", args)
```
yihui: I'd recommend not to quote `paste`, otherwise it can be unpredictable where this `paste` function could come from. If you do not quote it, R knows it is from the base package.

yihui:
I did not see any problems during my testing. I believe it is ready to go (after you fix the minor `paste` thing). Thanks!

yihui:
I just realized this PR was not merged. I rebased the branch and fixed the conflicts. @jcheng5 If you do not have anything else to add, could you merge this PR? Thanks!

PR #24, opened 2015-02-23
Do not insert newlines except at the start and end of a block

### LINKED ISSUE #19: Need some way of not printing whitespace
For example:

``` R
> div("foo", "bar")
<div>
  foo
  bar
</div> 
```

It would be good to be able to do something like one of the following:

``` R
div(.nospace = TRUE, "foo", "bar")
div("foo", nospace(), "bar")
nospace(div("foo", "bar"))
```

jimhester:
This is a simple fix for #19.  Previously it was impossible to fully control the white space as newlines were always added between items within a tag.  This made certain things impossible to construct e.g. `<h1>a<em>b</em></h1>` as the added newline would be included in the header so the `a` and `b` would be on different lines.

This implements @hadley's suggestion of only including newlines at the start and end of the block.

I updated all the tests to pass.  Let me know if this is useful and if so if you would like me to write some extra tests for it.

Closes #19

### DIFF
--- a/R/tags.R
+++ b/R/tags.R
@@ -455,7 +455,18 @@ doRenderTags <- function(x, indent = 0) {
   },
     finally = close(conn)
   )
-  return(HTML(paste(htmlResult, collapse = "\n")))
+  if (length(htmlResult) <= 2) {
+    return(HTML(paste(htmlResult, collapse = "\n")))
+  }
+  return(HTML(paste0(collapse = "",
+                     htmlResult[1],
+                     "\n",
+                     paste0(htmlResult[c(-1, -length(htmlResult))], collapse = ""),
+                     "\n",
+                     htmlResult[length(htmlResult)]
+                    )
+             )
+        )
 }
 
 # Walk a tree of tag objects, rewriting objects according to func.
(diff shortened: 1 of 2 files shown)


hadley:
Shouldn't this need a list of block-level tags?

jimhester:
Ah I misunderstood your earlier comment slightly this is more naive (and
wrong).  I think
https://developer.mozilla.org/en-US/docs/Web/HTML/Block-level_elements has
a complete list of block level elemnts, I will update the pull in a few
minutes.

jimhester:
With 3730c0f now explicitly testing for block level tags and only adding newlines to them.

jcheng5:
Soooo... this is a little embarrassing. It turns out I accidentally unwatched this repo, so I'm just now seeing this PR. Sorry.

I think something's still a little off:

```r
> renderTags(h1("a", em("b"), "c"), indent=1)$html
  <h1>
a      <em>b</em>c
  </h1> 
```

Don't we want that output to be
```r
  <h1>
    a<em>b</em>c
  </h1> 
```
?

jimhester:
I think this still needs more work to work properly, so I am going to close it for now.

PR #33, opened 2015-12-10
Add template support

wch:
This adds support for HTML templates, with the functions `htmlTemplate()` and `renderDocument()`.


R/template.R
```diff
@@ -0,0 +1,123 @@
...
+#'   be an object created by \code{\link{htmlDependency}}, or a list of such
+#'   objects. These dependencies will be added first, before other dependencies.
+#' @param processDep A function that takes a "raw" html_dependency object and
+#'   does further processing on it. For example, when \code{renderDocument} is
+#'   called from Shiny, the function \code{\link[shiny]{createWebDependency}} is
+#'   used; it modifies the href and tells Shiny to serve a particular path on
+#'   the filesystem.
+#'
+#' @export
+renderDocument <- function(x, deps = NULL, processDep = identity) {
```
wch: The one thing I don't really like is the `processDep` argument. When this function is called from shiny, it gets `processDep=createWebDependency`. That function modifies the `deps` a little bit, and it tells Shiny to serve up the paths. I didn't see a straightforward way to separate out that step, so I added the `processDep` argument.

R/template.R
```diff
@@ -0,0 +1,123 @@
...
+  })
+
+  # Create environment to evaluate code, as a child of the global env. This
+  # environment gets the ... arguments assigned as variables.
+  vars <- list(...)
+  if ("headContent" %in% names(vars)) {
+    stop("Can't use reserved argument name 'headContent'.")
+  }
+  vars$headContent <- function() HTML("<!-- HEAD_CONTENT -->")
+  env <- list2env(vars, parent = globalenv())
```
hadley: Shouldn't this `parent = parent.frame()`? Otherwise it will be annoying to call from another function

R/template.R
```diff
@@ -0,0 +1,123 @@
...
+
+
+#' Render an html_document object
+#'
+#' This function renders \code{html_document} objects, and returns a string with
+#' the final HTML content. It calls the \code{\link{renderTags}} function to
+#' convert any shiny.tag objects to HTML. It also finds any any web dependencies
+#' (created by \code{\link{htmlDependency}}) that are attached to the tags, and
+#' inserts those. To do the insertion, this function finds the string
+#' \code{"<!-- HEAD_CONTENT -->"} in the document, and replaces it with the web
```
hadley: I think you mean ` {{ headContent() }}` ?

NEWS
```diff
@@ -1,3 +1,10 @@
+htmltools 0.2.10
+--------------------------------------------------------------------------------
+
+* Add template support, with the `htmlTemplate()`, `renderDocument()`, and
+  `suppresDependencies()` functions.
```
robertzk: Typo `s/suppres/suppress`

R/template.R
```diff
@@ -0,0 +1,123 @@
...
+  pieces[-1] <- lapply(pieces[-1], function(piece) {
+    tagList(
+      eval(parse(text = piece[1]), env),
+      HTML(piece[[2]])
+    )
+  })
+
+  result <- tagList(pieces)
+
+  if (document_ == "auto"  ) {
```
robertzk: Any particular reasoning for the spacing after `"auto"`?

inst/tests/test-template.R
```diff
@@ -0,0 +1,61 @@
+context("templates")
+
+# Searches for an html dependency of format name[version], as in "d3[3.5.10]",
+# within the html-dependencies script tag
+findDep <- function(x, name, version) {
```
robertzk: Maybe place in `helper-findDep.R`? Are you anticipating any re-use in future tests?

robertzk:
I just wanted to pipe in that I think this is really neat.

R/template.R
jcheng5: Would that mean that local variables in the caller will become available to the template (i.e. dynamic scoping)? If so, we don't want that. We only want access to 1) the explicitly passed in vars (`...`), 2) the global environment, and 3) the search path.

jcheng5:
I just downloaded a random template (https://www.freewebsitetemplates.com/preview/zerotype/index.html) and dropped a plot and an action button into it, and it worked great!

jcheng5:
There does seem to be a problem with UTF-8 characters in templates, they're getting garbled for me (and I'm on Mac, I didn't even try Windows).

R/template.R
hadley: Oh, I get it.  Are you sure you want to make available variable from the global environment?  Maybe you want `parent.env(globalenv())` which will give everything in attached packages.

wch: That was actually how I originally implemented it, but there's a problem: if someone runs `library(foo)` in the template, that package won't actually be accessible in the template code.

R/template.R
wch: There's actually a two-step process: in the first step, `htmlTemplate()`, all the code in `{{ }}` blocks gets executed. The `headContent()` function gets replaced with `<!-- HEAD_CONTENT -->` in this stage. In the next step, `renderDocument()`, the `<!-- HEAD_CONTENT -->` gets replaced with header stuff. We discussed whether we wanted the `headContent()` function at all, or to just have users put in `<!-- HEAD_CONTENT -->`. We didn't have strong feelings either way, but decided to go with `headContent()`, which would also allow us to allow passing in arguments (like html dependencies) in the future.

R/template.R
```diff
@@ -0,0 +1,123 @@
...
+  deps <- lapply(deps, processDep)
+  depStr <- paste(sapply(deps, function(dep) {
+    sprintf("%s[%s]", dep$name, dep$version)
+  }), collapse = ";")
+  depHtml <- renderDependencies(deps, "href")
+
+  # Put content in the <head> section
+  head_content <- paste0(
+    '  <meta http-equiv="Content-Type" content="text/html; charset=utf-8"/>',
+    sprintf('  <script type="application/shiny-singletons">%s</script>',
```
wch: This part is also Shiny-specific... probably not harmful for non-Shiny use cases, but also not great to have hard-coded here.

R/template.R
wch: Typos - thanks for catching them. :-/

R/template.R
hadley: It seems like whatever you do it should be consistent between the docs and the example

R/template.R
hadley: Hmmm, I don't think it would be awful to disallow that. Having templates change global state is going to be a bad practice anyway

R/template.R
jcheng5: Maybe it's just the name that's a problem? How about `filterDependencies`?

R/template.R
jcheng5: I could be wrong but I think the shiny-singletons concept is actually htmltools now, it just inherited the name.

wch:
I added a test with UTF-8 characters that are 2, 3, and 4 bytes in the UTF-8 encoding, and it works fine. I also tried running a Shiny app using the same template, and it works fine. This was on Linux and Mac - I haven't tested Windows yet. Do you know what characters you used?

jcheng5:
OK, false alarm on UTF-8 problem. I added the characters `ððð` and they were rendered as `ÐÐÐ`, because the element I happened to drop them in was being uppercased by CSS. :)

jcheng5:
@ramnathv @timelyportfolio You guys may find this feature useful for making layouts to put widgets in.

PR #34, opened 2015-12-17
Add css function

jcheng5:


inst/tests/test-tags.r
```diff
@@ -570,6 +570,32 @@ test_that("Indenting can be controlled/suppressed", {
   )
 })
 
+test_that("cssList tests", {
+  expect_identical("", css())
+  expect_identical("", css())
+  expect_identical(
+    css(
+      font.family = 'Helvetica, "Segoe UI"',
```
aronatkins: If someone specifies `c('Helvetica','Segoe UI')`, should `css` know that certain values need quoting?

jcheng5: I'm choosing not to tackle any property-specific rules right now (I also won't wrap a URL in url() or append "px" to a number, etc.), and maybe not ever.

jcheng5: Actually you couldn't do two font families like that, because I join multiple values using " " not ",".

PR #35, opened 2015-12-18
Don't put newlines around inline tags

### LINKED ISSUE #19: Need some way of not printing whitespace

wch:
Fixes #19. This PR isn't ready yet but I wanted to put it up for some discussion before doing more on it.

One unequivocally good thing is that I switched from `readLines` to `readChar`, which improves performance a LOT when there are long lines. (I've run into this when serializing long JSON strings. Good thing we have nice profiling tools now :). For example, running this code with master vs. this branch:

```R
sizes <- round(runif(1e2, 1000, 100000))
strings <- lapply(sizes, function(size) {
  paste0(
    sample(c(letters, LETTERS), size, replace = TRUE),
    collapse = ""
  )
})

tag <- do.call(div, strings)

system.time({
  as.character(tag)
})

# master:
#    user  system elapsed 
#   0.560   0.000   0.562 

# New code:
#    user  system elapsed 
#   0.176   0.004   0.188 
```

If the strings are very short, there's no noticeable performance difference.

The next part is the actual change in output. The current changes satisfy a bunch of tests (at the end of this post) but introduces some possibly undesirable behavior. For example, consider the following:

```R
as.character(span("a", "b"))
```

On master, this renders as:

```html
<span>
  a
  b
</span>
```

On this branch, it renders as:

```html
<span>ab</span>
```

I would say in general that this is desirable, except for the fact that it could change the appearance of HTML from existing code that's out there. So we need to decide if that's what we really want to do.

Because of changes like this, the current PR breaks a lot of existing tests.

This PR does pass the following (not yet committed) tests:

```R
# Convert tag object to plain character string
textify <- function(x) {
  renderTags(x)$html[1]
}

expect_identical(
textify(div("x")),
"<div>x</div>"
)

expect_identical(
textify(span("x")),
"<span>x</span>"
)

expect_identical(
textify(div(span("x"))),
"<div>
  <span>x</span>
</div>"
)

expect_identical(
textify(span(div("x"))),
"<span>
  <div>x</div>
</span>"
)

expect_identical(
textify(div(div("x"))),
"<div>
  <div>x</div>
</div>"
)

expect_identical(
textify(span(span("x"))),
"<span><span>x</span></span>"
)

expect_identical(
textify(div(span("x"), span("x"))),
"<div>
  <span>x</span><span>x</span>
</div>"
)

expect_identical(
textify(span(div("x"), span("x"), span("x"))),
"<span>
  <div>x</div>
  <span>x</span><span>x</span>
</span>"
)

expect_identical(
textify(span(span("x"), div("x"), span("x"))),
"<span><span>x</span>
  <div>x</div>
  <span>x</span>
</span>"
)

expect_identical(
textify(div(span("x"), div("x"), span("x"))),
"<div>
  <span>x</span>
  <div>x</div>
  <span>x</span>
</div>"
)

expect_identical(
textify(div(span("x", div("x")))),
"<div>
  <span>x
    <div>x</div>
  </span>
</div>"
)
```

### DIFF
--- a/R/tags.R
+++ b/R/tags.R
@@ -502,19 +502,30 @@ doRenderTags <- function(x, indent = 0) {
   # since R doesn't have something like that (that I know of),
   # file() is the next best thing.
   conn <- file(open="w+b", encoding = "UTF-8")
+  # Track how many bytes we write, so we can read in the right amount
+  # later with readChar.
+  bytes <- 0
+
   connWriter <- function(text) {
-    text <- enc2utf8(text)
+    raw <- charToRaw(enc2utf8(text))
+    bytes <<- bytes + length(raw)
     # This is actually writing UTF-8 bytes, not chars
-    writeBin(charToRaw(text), conn)
+    writeBin(raw, conn)
   }
-  htmlResult <- tryCatch({
-    tagWrite(x, connWriter, indent)
-    flush(conn)
-    readLines(conn, encoding = "UTF-8")
-  },
+
+  htmlResult <- tryCatch(
+    {
+      tagWrite(x, connWriter, indent)
+      flush(conn)
+
+      # Strip off trailing \n (which is always there) but make sure not to
+      # specify a negative number of chars.
+      bytes <- max(bytes - 1, 0)
+      readChar(conn, bytes, useBytes = TRUE)
+    },
     finally = close(conn)
   )
-  return(HTML(paste(htmlResult, collapse = "\n")))
+  return(HTML(htmlResult))
 }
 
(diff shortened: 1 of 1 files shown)


jcheng5:
After (another) long discussion about this with @wch, we've decided the first step is to add an explicit argument to tag functions that indicate whether whitespace should be added on the left and/or right.

PR #36, opened 2015-12-30
Always overwrite dependency files when copying them

yihui:
Also changed file.exists() to dir_exists() for the reason described at https://github.com/rstudio/rmarkdown/pull/347

### DIFF
--- a/R/html_dependency.R
+++ b/R/html_dependency.R
@@ -237,20 +237,19 @@ copyDependencyToDir <- function(dependency, outputDir, mustWork = TRUE) {
   target_dir <- file.path(outputDir,
     paste(dependency$name, dependency$version, sep = "-"))
 
-  if (!dir_exists(target_dir)) {
+  if (!dir_exists(target_dir))
     dir.create(target_dir)
 
-    srcfiles <- file.path(dir, list.files(dir))
-    destfiles <- file.path(target_dir, list.files(dir))
-    isdir <- file.info(srcfiles)$isdir
-    destfiles <- ifelse(isdir, dirname(destfiles), destfiles)
+  srcfiles <- file.path(dir, list.files(dir))
+  destfiles <- file.path(target_dir, list.files(dir))
+  isdir <- file.info(srcfiles)$isdir
+  destfiles <- ifelse(isdir, dirname(destfiles), destfiles)
 
-    mapply(function(from, to, isdir) {
-      if (isdir && !dir_exists(to))
-        dir.create(to)
-      file.copy(from, to, overwrite = TRUE, recursive = TRUE)
-    }, srcfiles, destfiles, isdir)
-  }
+  mapply(function(from, to, isdir) {
+    if (isdir && !dir_exists(to))
+      dir.create(to)
+    file.copy(from, to, overwrite = TRUE, recursive = isdir)
+  }, srcfiles, destfiles, isdir)
 
   dependency$src$file <- normalizePath(target_dir, "/", TRUE)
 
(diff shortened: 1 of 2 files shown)


jjallaire:
This looks okay to me. Given that this relates to removing/overwriting files perhaps we should have one more pair of eyes reviewing it. @jcheng5 or @wch  ?

R/html_dependency.R
```diff
@@ -231,32 +231,37 @@ copyDependencyToDir <- function(dependency, outputDir, mustWork = TRUE) {
     }
   }
 
-  if (!file.exists(outputDir))
+  if (!dir_exists(outputDir))
     dir.create(outputDir)
```
hadley: A simpler alternative is to remove the existence check and do `dir.create(outputDir, showWarnings = FALSE)`

R/html_dependency.R
```diff
@@ -231,32 +231,37 @@ copyDependencyToDir <- function(dependency, outputDir, mustWork = TRUE) {
...
 
   target_dir <- file.path(outputDir,
     paste(dependency$name, dependency$version, sep = "-"))
 
-  if (!file.exists(target_dir)) {
-    dir.create(target_dir)
+  if (dir_exists(target_dir))
+    unlink(target_dir, recursive = TRUE)
```
hadley: This scares me - if you accidentally specify the wrong target, you're going to wipe out a directory

wch: I think it's relatively safe though, since `target_dir` is generated from `outputDir` and the dependency name and version.

hadley: But why is it necessary at all? Because R doesn't have a decent recursive copy function?

wch:
Looks good to me.

R/html_dependency.R
yihui: The reason for this was described in the commit message 480917c1989bc601e9e3f12676055d6fed8de30e Basically we don't there are leftover files in the target dependency dir, e.g. you may have lib/foo.js last time, and it was removed from the original library, then the next time you copy the library over to the target dir, you want to remove this lib/foo.js as well. It won't be removed only with `file.copy(recursive = TRUE)`.

I'll add a comment to explain it to the R code here.

R/html_dependency.R
yihui: I'm not sure if this suppresses other possible warnings as well. If it does, I'd rather stick to the current approach.

R/html_dependency.R
```diff
@@ -231,32 +227,39 @@ copyDependencyToDir <- function(dependency, outputDir, mustWork = TRUE) {
...
   target_dir <- file.path(outputDir,
     paste(dependency$name, dependency$version, sep = "-"))
 
-  if (!file.exists(target_dir)) {
-    dir.create(target_dir)
+  # completely remove the target dir because we don't want possible leftover
+  # files in the target dir, e.g. we may have lib/foo.js last time, and it was
+  # removed from the original library, then the next time we copy the library
+  # over to the target dir, we want to remove this lib/foo.js as well
+  if (dir_exists(target_dir)) unlink(target_dir, recursive = TRUE)
```
jcheng5: `unlink(recursive = TRUE)` makes my stomach hurt... just for peace of mind, can we do some sanity checks here, e.g. make sure that dependency$name and dependency$version are length 1 and not `""`, and make sure `outputDir` is non-`""`, non-`"/"`, etc.?

I'm sure we can prove that this particular snapshot of the code is correct, but more checks and comments with exclamation points will help protect us as the code evolves over time.

https://www.google.com/webhp?hl=en#hl=en&q=rm+rf+bug

yihui: Sure. I'll add some sanity checks soon.

PR #38, opened 2016-01-01
Sanity checks for dependency name and version in htmlDependency(), as well as outputDir in copyDependencyToDir()

yihui:
I checked `/` and `\` because we don't want something like `htmlDependency(name = '../../home')`.

Cc @jcheng5

### DIFF
--- a/R/html_dependency.R
+++ b/R/html_dependency.R
@@ -74,6 +74,10 @@ htmlDependency <- function(name,
     )
   }
 
+  version <- as.character(version)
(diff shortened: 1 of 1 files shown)


jjallaire:
@jcheng5 You can merge this if you think it provides adequate safeguards. If we are really spooked about this we could just to a copy recursive without the unlink. This would create a risk of leaving leftover/unused files in the directory if a same version updated of a dependency actually *removed* files (seems like that would be rare).

PR #39, opened 2016-01-01
Move tests to the standard place

yihui:


wch:
I ended up doing this manually because some files changed resulting in a merge conflict.

PR #42, opened 2016-01-21
Allow html dependencies to be added inline

wch:
This PR allows html dependencies to be added inline, as a child of a tag, or as an element of a `tagList`. For example, these all work now:

```R
dep <- htmlDependency("jqueryui", "1.11.4", c(href="shared/jqueryui"),
                      script = "jquery-ui.min.js")

# Inline as a child of the div()
div("Code here", dep)

# Inline in a tagList
tagList(div("Code here"), dep)

# With attachDependencies (the existing way to do it)
attachDependencies(div("Code here"), dep)
```

The key change was to add an `as.tags.html_dependency` method, which simply creates an empty `tagList()` and attaches the dependency to it; then `as.tags` needs to be called on a tag's children when the tag is created, or when children are modified.

There was opportunity for greatly simplifying `tagify`, `flattenTags`, `rewriteTags`, and `tagWrite` by using S3, so I did that as well.

Some internal data structures have changed, which resulted in a change in some singleton hashes. This is because a tag or tagList now never contains plain lists; `as.tags` is called recursively on all children (or tagList elements), and this converts lists to `tagList` objects.

Note that there's a little clunkiness in some `as.tags` methods to preserve the singleton attribute. I have a separate branch where singletons are implemented as a class instead of attribute, resulting in somewhat more elegant code. There are a couple pros and cons to that, which is why I didn't add it to this PR (I can make a new PR separately after this one is merged).


jcheng5:
Shouldn't `as.tags.html_dependency` alone, without any of the other changes, have worked? Did it matter that `as.tags` was called at render time rather than at construction time? I wouldn't assume that that change won't break anything--we are doing some [crazy stuff in as.tags in shiny](https://github.com/rstudio/shiny/blob/12b0484e9a74dea0a1c32b47b231487ad6734ffe/R/shinywrappers.R#L24-L39).

wch:
Hm, it does look like it'll work to call `as.tags` at render time, although `as.tags` will also need to be called in `findDependencies`.

PR #44, opened 2016-01-22
as.character.shiny.tag returns character. Closes #31

### LINKED ISSUE #31: as.character doesn't convert the tag into a character to rendering
as.character should return the tag as a character. but it doesn't

```R
tags$b("City")
<b>City</b> 
as.character(tags$b("City"))
<b>City</b> 
```

htmltools v0.2.6
RRO v 3.2.1
Windows 8.1

wch:
Before this fix:

```
> str(as.character(div("x")))
Classes 'html', 'character'  atomic [1:1] <div>x</div>
  ..- attr(*, "html")= logi TRUE
```

After:

```
> str(as.character(div("x")))
 chr "<div>x</div>"
```

### DIFF
--- a/R/tags.R
+++ b/R/tags.R
@@ -180,7 +180,7 @@ format.shiny.tag <- function(x, ..., singletons = character(0), indent = 0) {
 
 #' @export
 as.character.shiny.tag <- function(x, ...) {
-  renderTags(x)$html
+  as.character(renderTags(x)$html)
 }
 
 #' @export


wch:
Hm, this breaks lots of tests, because the tests expect something like this:

```R
expect_identical(as.character(tagList("hi")), HTML("hi"))
```

It might be worth having a discussion about what the desired behavior is.

jcheng5:
I prefer the behavior we have today. I did that on purpose so that you could put an `as.character(tags)` inside of another list or tag and get the right results (i.e. no double-encoding). We know it's HTML, why not mark it as such?

wch:
Two reasons:

* `as.character` (and similar) almost always return a plain vector, without extra classes or attributes.
* Right now, if you want a plain string (character without extra classes), you need to run `as.character(as.character(tags))`. It seems like `as.character()` should be idempotent.

But if changing it will break things, then maybe it's not worth it.

jcheng5:
Ok. The docs for as.character do say that attributes will be stripped, including names. I guess it's better for us to swallow the pill now, at the start of the cycle.

PR #47, opened 2016-02-05
Fix handling of brackets in templates

### LINKED ISSUE #46: Error if template block has no leading or trailing spaces
For example:

```R
htmlTemplate(text_= "{{ 1+1 }}")
# Error in htmlTemplate(text_ = "{{ 1+1 }}") : 
#   Mismatched {{ and }} in HTML template.

htmlTemplate(text_= " {{ 1+1 }}")
# Error in FUN(X[[i]], ...) : Mismatched {{ and }} in HTML template. 

htmlTemplate(text_= "{{ 1+1 }} ")
# Error in htmlTemplate(text_ = "{{ 1+1 }} ") : 
#   Mismatched {{ and }} in HTML template.

htmlTemplate(text_= " {{ 1+1 }} ")
# [1] " \n2\n "
```

wch:
This fixes #46. It implements a DFA in C++ for the template handling.

Note that this introduces an Rcpp dependency, but I think that's not much of a burden because almost everyone will already have Rcpp installed (they will have installed Shiny, which uses httpuv, which uses Rcpp).

Notes:
* I can squash the commits if that's preferable.
* I'd welcome any feedback on the C++ code.


src/template.cpp
```diff
@@ -0,0 +1,146 @@
+#include <Rcpp.h>
+using namespace Rcpp;
+
+// Positive substring: return a substring of `str``going from `start` to `end`;
+// if `end` is earlier than `start`, return empty string.
+std::string pos_substr(std::string str, int start, int end) {
```
jcheng5: FWIW, I think it's better to treat `start` as inclusive and `end` as exclusive. This has a number of nice benefits over having an inclusive `end`:

- The length of the substring is just `end - start`
- You can move onto the next piece by using the old `end` value as the new `start` value
- It's possible to get a zero length string by having `start == end`; you generally don't get into `end < start` situations like you do here

src/template.cpp
```diff
@@ -0,0 +1,146 @@
...
+      switch (c) {
+      case '}':
+        state = code_oneCloseBracket; break;
+      case '\'':
+        state = code_string1; break;
+      case '"':
+        state = code_string2; break;
+      case '`':
+        state = code_backtick; break;
+      case '\\':
```
jcheng5: What does `\` mean outside of a string?

src/template.cpp
jjallaire: +1

src/template.cpp
```diff
@@ -0,0 +1,146 @@
...
+  if (end < start)
+    return "";
+  else
+    return str.substr(start, end - start + 1);
+}
+
+// Break template text into character vector. The first element element of the
+// resulting vector is HTML, the next is R code, and they continue alternating.
+// [[Rcpp::export]]
+CharacterVector template_dfa(CharacterVector x) {
```
jjallaire: You can just declare the parameter as std::string if that's what you want to ultimately handle (the Rcpp::as will be done behind the scenes)

src/template.cpp
```diff
@@ -0,0 +1,146 @@
...
+    code_backtick,
+    code_backtick_backslash,
+    code_backslash
+  };
+
+  if (x.length() != 1) {
+    stop("Input HTML must be a character vector of length 1");
+  }
+  std::string input = Rcpp::as<std::string>(x[0]);
+  CharacterVector pieces(0);
```
jjallaire: You should declare this as std::vector<std::string> rather than CharacterVector. The reason is that since vectors are immutable in R when you do a `push_back` on a CharacterVector it's actually re-allocating an entirely new vector from scratch and copying the old elements into it! If you use std::vector<std::string> then you'll get nicer re-allocation behavior and you can just return it and Rcpp will do Rcpp::wrap on it automatically (turning it into a CharacterVector).

jjallaire:
There is *one* issue created by the Rcpp dependency here. The automatic installation of htmltools in RStudio (done as part of installing rmarkdown or shiny on demand) specifies type = "source". This is because the CRAN Mac build servers have been so far behind (and in many cases remiss in building for some platforms). So if we do this then I think RStudio needs to stop using type = "source" for htmltools (unfortunately we can't undo that for binaries still in the field, but I can make the change now to limit the damage).

src/template.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+  int i = 0;
+  int pieceStartIdx = 0;
+  int len = input.length();
+  char c;
+  State state = html;
+  while (i < len) {
```
jcheng5: You don't have to change this, but, I tend to use a for-loop instead of while-loop, to make it harder to accidentally bypass `i++` (for example, a `continue` statement anywhere below--you don't have any but it can sometimes be useful).

jcheng5:
@jjallaire Is `type="source"` going to be a problem just for relatively recent versions of RStudio? Or have you been installing htmltools this way for a long time?

jjallaire:
It looks like it's quite recent (December of 2015) so the impact is minimal. I'll change it back right now and do a patch release and then I don't think we need to be worried about this at all.

jcheng5:
Great, thanks!

jcheng5:
- Technically `%}}%` is valid R code :)
- What about R comments? This currently doesn't work, should it?  
`htmltools::htmlTemplate(text_ = "foo {{ 'hi' # that's cool }} bar")`

src/template.cpp
wch: Oh yeah, that makes a lot more sense here. In my initial R implementation it made sense to use a `while` loop (to avoid allocating the integer vector for counting) but a `for` loop would be better here.

src/template.cpp
wch: Hm, nothing that I can think of now. I'll take it out.

wch:
Thanks for the feedback guys! I've implemented the changes you suggested.

As for comments, you're right, and I've added support for them.

PR #48, opened 2016-02-07
Make it possible to copy only the files listed in script/stylesheet/attachment of htmlDependency()

yihui:
This is to provide a straightfoward solution to https://github.com/ramnathv/htmlwidgets/pull/136, i.e. `htmlDependency(all_files = FALSE)`. And more importantly, make code chunks that contain HTML widgets cacheable in **knitr**. Currently they are not cacheable only because of the hack in htmlwidgets (copy the js binding to temp dir, and temp dir does not persist between R sessions).

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -7,8 +7,7 @@ htmltools 0.3.2
 * `htmlTemplate` now can use a string as a template instead of requiring a
   file. (#41, #43)
 
-htmltools 0.3.1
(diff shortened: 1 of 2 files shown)


jcheng5:
LGTM. Nice improvement. @wch?

wch:
Looks good to me too.

PR #53, opened 2016-03-18
Allow dependencies to be merged when resolved

yihui:
Added a `mergeable` argument to `htmlDependency()`.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -14,6 +14,10 @@ htmltools 0.3.5
   files under the src directory should be copied when rendering dependencies,
   or only those specified in the dependency objects. (#48)
 
+* `htmlDependency()` gained a new argument `mergeable` to specify whether one
+  dependency can be merged with another of the same name, version, and src, when
+  the dependencies are to be resolved. (#53)
+
 * `copyDependencyToDir()` will always completely overwrite the target directory
   when copying HTML dependency files to make sure all dependency files are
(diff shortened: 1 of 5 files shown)


jjallaire:
LGTM. @wch and @jcheng5 could you also take a look?

wch:
Looks good to me.

jcheng5:
@yihui Can you explain what this is for?

yihui:
@jcheng5 Sure. Here is the motivation: https://github.com/rstudio/rmarkdown/issues/645 For example, when using the Buttons extension in DT, users may choose a subset of buttons, and different buttons require different JS modules: https://github.com/rstudio/DT/tree/master/inst/htmlwidgets/lib/datatables-extensions/Buttons/js Some modules are pretty big (e.g. the ones for exporting the table to PDF or Excel), so I do not want to include all scripts on the page. If you have two HTML dependencies that have the same name and version but different JS files, the second dependency will simply be dropped. So if I have two widgets that use different Buttons modules, the second widget will not render properly.

jcheng5:
@yihui Can you make those distinct dependencies (with differing names)? That's the approach we've taken in Leaflet.

yihui:
@jcheng5 I'm going to split the dependencies in DT into smaller units, so I'll close this PR for now, and please feel free to make a CRAN release with the current master branch. Thanks!

PR #54, opened 2016-03-18
Print tags as HTML instead of character

yihui:
This is an issue introduced by #44.

For example, `tags$a(href = "#", "link")` should be printed as

```html
<a href="#">link</a>
```

instead of

```
[1] "<a href=\"#\">link</a>"
```


wch:
LGTM

PR #56, opened 2016-04-12
Add viewport units (vw, vh, vmin, vmax) as valid CSS units

mkuhn:
For specifying the dimensions (especially the height) of figures in Shiny apps it's very useful to access the viewport height. Specifying the height as pixels is too restrictive, and as percent is insufficient: the surrounding `div` may have a small size, and will not increase its size to the height of the browser window. Luckily, there are the viewport-based CSS units, which should pass validation. 

Sources: http://stackoverflow.com/questions/1575141/make-div-100-height-of-browser-window http://caniuse.com/#feat=viewport-units

### DIFF
--- a/R/tags.R
+++ b/R/tags.R
@@ -1380,7 +1380,7 @@ validateCssUnit <- function(x) {
     x <- as.numeric(x)
 
   pattern <-
-    "^(auto|inherit|((\\.\\d+)|(\\d+(\\.\\d+)?))(%|in|cm|mm|em|ex|pt|pc|px|vh))$"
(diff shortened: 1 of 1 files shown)


wch:
Great, can you update the documentation for the function, and also submit a contributor agreement?
https://www.rstudio.com/wp-content/uploads/2014/06/RStudioIndividualContributorAgreement.pdf

If that's too complicated, we could just re-implement the same feature.

mkuhn:
As I also have two pull requests at d3heatmap, I just sent in the contributor agreement.

Issue #59, opened 2016-04-28
consider annotating dependency information


kevinushey:
It would be useful if [`renderDependencies()`](https://github.com/rstudio/htmltools/blob/master/R/html_dependency.R#L375-L464) annotated the injected `<script>` / `<link>` tags with data attributes describing the source of each dependency, e.g.

```
 <script
        data-name="htmlwidgets"
        data-version="0.6"
        data-script="htmlwidgets.js"
        data-stylesheet=""
        src="/path/to/R/library/htmlwidgets/www/htmlwidgets.js"></script>
```

This would make it possible to recover the source of a dependency after e.g. `rmarkdown::render()` is done with the document.

jcheng5:
I'm a little leery of putting the user library's absolute path in the html output, if the latter may be published somewhere. Can this be constrained to cases where we know the output isn't going anywhere?

kevinushey:
Sorry, I should have been more clear -- all I'm advocating for is the inclusion of the `data-*` attributes, which should be easy to draw from the dependency information already passed into `renderDependencies()`. (The `src` field would just be whatever is normally passed in)

yihui:
Actually I wished the dependency object had a `package` attribute, and `src` could be resolved at a later time using `system.file(src, package = package)` (if the `package` name is provided). This will hide the absolute path, and make the dependencies more portable since they will be independent of the R library path.

jjallaire:
I agree. I think this could be pretty seamlessly added (if there is a
`package` attribute then `src` can be relative to the package).

PR #60, opened 2016-05-03
Add a package argument to htmlDependency()

yihui:
As mentioned in #59, I think it will be nice to provide a `package` argument in htmlDependency() so that `src` is resolved to absolute paths later. One benefit of this is that knitr can cache the dependencies without having to store absolute paths.


jjallaire:
I think this is great. Too bad we can't retroactively get all the widgets in the wild to use it!

PR #68, opened 2016-09-08
Ensure dependencies are writeable

rekado:
Rmarkdown copies dependencies to a temporary directory and tries to remove the directory afterwards.  When a dependency's file mode was set to 444 at package installation (as is the case on Guix), the copied file cannot be removed by R.

Setting `copy.mode` to `FALSE` ensures that the files can be removed by the user (or by the same R process at a later point).

### DIFF
--- a/R/html_dependency.R
+++ b/R/html_dependency.R
@@ -315,7 +315,7 @@ copyDependencyToDir <- function(dependency, outputDir, mustWork = TRUE) {
       dir.create(dirname(to), recursive = TRUE)
     if (isdir && !dir_exists(to))
       dir.create(to)
-    file.copy(from, to, overwrite = TRUE, recursive = isdir)
+    file.copy(from, to, overwrite = TRUE, recursive = isdir, copy.mode = FALSE)
   }, srcfiles, destfiles, isdir)
 
   dependency$src$file <- normalizePath(target_dir, "/", TRUE)


yihui:
Looks good to me.

wch:
@rekado Have you signed the RStudio contributor agreement?
https://www.rstudio.com/wp-content/uploads/2014/06/RStudioIndividualContributorAgreement.pdf

rekado:
Hi Winston,

> @rekado Have you signed the RStudio contributor agreement?
> https://www.rstudio.com/wp-content/uploads/2014/06/RStudioIndividualContributorAgreement.pdf

no, I have not.  (I think this contribution is minor enough to not count
as copyrightable.)

Does the agreement only apply to contributions that I specifically send
upstream to RStudio projects, and not to those that I make to RStudio
packages without sending upstream?

~~ Ricardo

yihui:
Yeah, for such a minor change, it is probably not necessary to sign the agreement. JJ has merged the similar PR in rmarkdown https://github.com/rstudio/rmarkdown/pull/800

jjallaire:
Agreed we don't need the agreement for this magnitude of change. We can
merge right away. Ricardo, for future reference if you do ever sign the
agreement for another change. it only applies to changes which you send
upstream.

rekado:
JJ Allaire <notifications@github.com> writes:

> Agreed we don't need the agreement for this magnitude of change. We can
> merge right away. Ricardo, for future reference if you do ever sign the
> agreement for another change. it only applies to changes which you send
> upstream.

Thanks for the clarification!

~~ Ricardo

rekado:
Hi there,

I was just bitten by this problem and thought to myself that this sounds awfully familiar... As I was about to send you a patch I found this!

Are there any obstacles to merging this?

~~ Ricardo

PR #71, opened 2016-09-30
Fix encoding issue on Windows

### LINKED ISSUE #69: Problem with encoding on Windows if head is UTF-8 but body is ASCII
See rstudio/shiny#1395.

We also need to test this when the HEAD contains characters that are in Windows-1252 but not UTF-8, like accented latin characters.

wch:
This fixes #69.

### DIFF
--- a/tests/testthat/test-template.R
+++ b/tests/testthat/test-template.R
@@ -46,6 +46,27 @@ test_that("UTF-8 characters in templates", {
   expect_identical(charToRaw(text), as.raw(c(0xc3, 0xbf)))
 })
 
+test_that("UTF-8 characters in template head but not body", {
+  # On Windows, a string with "中文" will automatically be marked as UTF-8.
+  ui <- tagList(
+    tags$head(tags$script("alert('中文')")),
+    "test"
+  )
+  html <- htmlTemplate("template-basic.html", body = ui)
+  res <- renderDocument(html)
+  expect_identical(Encoding(res), "UTF-8")
+  expect_true(grepl("中文", res, fixed = TRUE))
+
+  # On Windows, a string with "á" will automatically be marked as latin1.
+  ui <- tagList(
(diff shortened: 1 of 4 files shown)


tests/testthat/test-template.R
```diff
@@ -46,6 +46,27 @@ test_that("UTF-8 characters in templates", {
   expect_identical(charToRaw(text), as.raw(c(0xc3, 0xbf)))
 })
 
+test_that("UTF-8 characters in template head but not body", {
+  # On Windows, a string with "中文" will automatically be marked as UTF-8.
+  ui <- tagList(
+    tags$head(tags$script("alert('中文')")),
```
yihui: I just realized these tests might fail on Windows. I'm not sure if R can read these tests with the correct encoding, especially when the system locale is not Chinese. We may want to `skip_on_cran()`.

wch: My Windows VM definitely doesn't have its locale set to Chinese. But it was using a recent version of Windows; I agree that it's possible that older versions of windows might not handle this the same way. I'd prefer using `rawToChar()` instead (like in the previous test) instead of skipping the test.

It makes sense in this case to explicitly mark the string as UTF-8 on all platforms, but for the next test (with `á`) I'm not sure what the right thing to do is. On Unixes, the string would be automatically marked as UTF-8, but on Windows, it gets marked as latin1 (at least it did on my VM).

wch: FWIW, I checked the package on r-hub (Windows 2008 Server, R 3.3.2) and there were no errors. I assume that the locale isn't set to Chinese there.
https://builder.r-hub.io/status/htmltools_0.3.6.tar.gz-93b5e051f7594422861bf773c32de71f

yihui: Okay great!

PR #72, opened 2016-10-20
experiment for JSX

timelyportfolio:
This pull is solely intended as a point of discussion.  It would be really nice (but probably an edge case) to be able to produce `JSX`.  In `JSX` attributes are often unquoted JavaScript.  To achieve this in a naive way with ugly code, I made the changes in this pull.  I am willing to work much harder on this, but I want to make sure it is acceptable before I expend the effort.

### Example

I am working on the [`reactR`](https://github.com/timelyportfolio/reactR) package that would ease working with `React`/`JSX` in R.  As I was experimenting, since `tags` does not allow unquoted attributes, I currently have to do something like [this](https://github.com/timelyportfolio/react-micro-bar-chart/blob/master/code.R).

```
library(htmltools)
library(reactR)

attachDependencies(
  browsable(
    tagList(
      tags$head(tags$script(src="//d3js.org/d3.v3.min.js")),
      tags$body(tags$script(babel_transform(
'
ReactDOM.render(
  <MicroBarChart data={[1,5,2,4]} hoverColor="rgb(161,130,214)"
  fillColor="rgb(210,193,237)" />,
  document.body
)
'
      )))
    )
  ),
  list(
    html_dependency_react(),
    htmlDependency(
      name="microbarchart",
      version="0.1",
      src=file.path(getwd(),"src"),
      script="react-micro-bar-chart.js"
    )
  )
)
```

It would be really nice since I am so spoiled by `htmltools` to do this instead.

```
htmltools::tag("MicroBarChart", list(data=noquote('{[1,5,2,4]}'), hoverColor="rgb(161,130,214)", fillColor="rgb(210,193,237)"))
```

which after this pull gives us.

```
<MicroBarChart data={[1,5,2,4]} hoverColor="rgb(161,130,214)" fillColor="rgb(210,193,237)"></MicroBarChart>
```

### DIFF
--- a/R/tags.R
+++ b/R/tags.R
@@ -395,7 +395,13 @@ tagWrite <- function(tag, textWriter, indent=0, eol = "\n") {
   textWriter(paste8(indentText, "<", tag$name, sep=""))
 
   # Convert all attribs to chars explicitly; prevents us from messing up factors
-  attribs <- lapply(tag$attribs, as.character)
+  attribs <- lapply(tag$attribs, function(x){
+    if(inherits(x, c("AsIs","noquote"))){
+      x
+    } else {
+      as.character(x)
+    }
+  })
   # concatenate attributes
   # split() is very slow, so avoid it if possible
   if (anyDuplicated(names(attribs)))
@@ -404,11 +410,14 @@ tagWrite <- function(tag, textWriter, indent=0, eol = "\n") {
   # write attributes
   for (attrib in names(attribs)) {
     attribValue <- attribs[[attrib]]
-    if (!is.na(attribValue)) {
+    if (!is.na(attribValue) && !inherits(attribValue,c("AsIs","noquote"))) {
       if (is.logical(attribValue))
         attribValue <- tolower(attribValue)
       text <- htmlEscape(attribValue, attribute=TRUE)
       textWriter(paste8(" ", attrib,"=\"", text, "\"", sep=""))
+    }
+    else if(inherits(attribValue,c("AsIs","noquote"))){
+      textWriter(paste8(" ", attrib, "=", attribValue, sep="") )
     }
     else {
       textWriter(paste8(" ", attrib, sep=""))


timelyportfolio:
ping to see if there are any thoughts regarding this idea

jcheng5:
Sorry this slipped through the net the first time. I like it. I'd maybe change the function/class name from `noquote` to something more unique, like `rawattrib` or something? `noquote` sounds like something out of the rlang package.

I'd also refactor the code slightly so we test for this case before testing for `!is.na(attribValue)`, so you don't have to repeat the test twice.

@wch, any thoughts?

timelyportfolio:
@jcheng5 no problem at all, I had considered it a long shot anyways but it sounds like maybe there is more chance than I thought.  I chose `noquote` since it is built-in to R core, it implies what a user wants (in a different context though), and means we don't have to manage a new `class` or `object`.  If you think another way of handling is better or more intuitive, I am happy to change.  Also, makes sense on refactoring and will change after @wch comments.

Thanks!!!!!!!

timelyportfolio:
any thoughts no this especially now that `reactR` is more viable for `htmlwidgets` and `Shiny` inputs?

timelyportfolio:
one last effort - any hope?

cpsievert:
I think I agree with @jcheng5 on this one. It seems safer to introduce a new `rawattrib()` (or similar) class to avoid the chance of breaking any existing code. I think we'd be willing to take it on with that in mind (we should also do the check once and include unit test(s))

jcheng5:
How about `HTML()`, as suggested in https://github.com/rstudio/htmltools/issues/17?

timelyportfolio:
@cpsievert @jcheng5 thanks so much for the consideration.  Given the lack of any other user expression of interest or desire, I think better to just close this out and not try to squeeze `JSX` into `HTML`.

Really appreciate all the effort on `htmltools`.

PR #76, opened 2017-02-03
Add after_widget argument

cpsievert:
This adds more explicit control over dependency ordering for htmlwidget authors. See also https://github.com/ramnathv/htmlwidgets/pull/259


jcheng5:
You should be able to do this purely in htmlwidgets, there's no need to introduce the concept of a widget at this level. I'll elaborate in the htmlwidget PR.

PR #85, opened 2017-06-20
`list` argument for `tagAppendAttributes`

nteetor:
Similar to how `tagAppendChildren` allows passing elements through `...` or the `list` argument, it would be useful if `tagAppendAttributes` accepted a list of attributes through a `list` argument in addition to `...`.


cpsievert:
I don't think we'll be taking this on since you can now use `!!!` to splice in a list

```r
tagAppendAttributes(div(), !!!list(a = 1))
#> <div a="1"></div>
```

PR #87, opened 2017-06-28
Import option for htmlDependency

briangottfried:
Allow html dependencies to have an import option (i.e, `<link rel="import" href="..."></link>`). This is useful for importing custom web components.


cpsievert:
I don't think we'll be taking this on since this is not widely implemented/obsolete https://developer.mozilla.org/en-US/docs/Web/Web_Components/HTML_Imports

PR #95, opened 2018-04-25
`as.character.shiny.tags()` will handle non-ASCII attributes correctly

shrektan:
## Description

- `as.character.shiny.tags()` will handle non-ASCII attributes correctly if they are not encoded in native encoding.
- It's especially problematic on Windows when you need to use non-ASCII strings as the attributes but the strings are encoded in UTF-8.

### Before this PR

```r
library(htmltools)
latin1_str2 <- rawToChar(as.raw(c(255, 13, 10)))
Encoding(latin1_str2) <- "latin1"
as.character(tags$span(latin1_str2, title = latin1_str2))
#> [1] "<span title=\"<ff>&#13;&#10;\">ÿ\r\n</span>"
```

### After this PR

```r
library(htmltools)
latin1_str2 <- rawToChar(as.raw(c(255, 13, 10)))
Encoding(latin1_str2) <- "latin1"
as.character(tags$span(latin1_str2, title = latin1_str2))
#> [1] "<span title=\"ÿ&#13;&#10;\">ÿ\r\n</span>"
```

### DIFF
--- a/tests/testthat/test-tags.r
+++ b/tests/testthat/test-tags.r
@@ -681,9 +681,15 @@ test_that("Latin1 and system encoding are converted to UTF-8", {
   # ensure the latin1 attribute returns correctly after escaping
   latin1_str2 <- rawToChar(as.raw(c(0xff, 0x0d, 0x0a)))
   Encoding(latin1_str2) <- "latin1"
+  spanLatin <- as.character(tags$span(latin1_str2, title = latin1_str2))
+  expect_identical(Encoding(spanLatin), "UTF-8")
   expect_identical(
-    as.character(tags$span(latin1_str2, title = latin1_str2)),
-    "<span title=\"\u00ff&#13;&#10;\">\u00ff\r\n</span>"
+    charToRaw(spanLatin),
+    as.raw(c(0x3c, 0x73, 0x70, 0x61, 0x6e, 0x20, 0x74, 0x69, 0x74,
+             0x6c, 0x65, 0x3d, 0x22, 0xc3, 0xbf, 0x26, 0x23, 0x31, 0x33, 0x3b,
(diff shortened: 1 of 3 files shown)


shrektan:
Haha. Yes, just find that we need to mark the `Encoding` to `UTF-8` in order to work on a Windows machine. I've force-pushed it.

tests/testthat/test-tags.r
```diff
@@ -677,6 +677,14 @@ test_that("Latin1 and system encoding are converted to UTF-8", {
 
   expect_identical(Encoding(format(HTML(latin1_str))), "UTF-8")
   expect_identical(Encoding(format(tagList(latin1_str))), "UTF-8")
+
+  # ensure the latin1 attribute returns correctly after escaping
+  latin1_str2 <- rawToChar(as.raw(c(0xff, 0x0d, 0x0a)))
+  Encoding(latin1_str2) <- "latin1"
+  expect_identical(
```
wch: Can you change these tests to use the same pattern as above? That is, use `charToRaw()` and then compare the raw output to a specific byte sequence. Also, test that the encoding is UTF-8.

The reason it's important to do tests this way is because Unicode escaped values like `\u00ff` can be encoded in different ways, so simply using a string like `"\u00ff"` does not specify a byte sequence. Also, `expect_identical()` uses `identical()`, which automatically does encoding conversions. For example, on my Mac, I can run this:


```R
x <- rawToChar(as.raw(c(0xc3, 0xbf)))
Encoding(x) <- "UTF-8"

x
#> [1] "ÿ"
Encoding(x)
#> [1] "UTF-8"
charToRaw(x)
#> [1] c3 bf

x1 <- iconv(x, to="latin1")
x1
#> [1] "ÿ"
Encoding(x1)
#> [1] "latin1"
charToRaw(x1)
#> [1] ff

# This passes (!)
expect_identical(x, x1)
```

shrektan: Thanks for sharing the reason! I've changed the test.

PR #100, opened 2018-06-12
Make sure NA attributes are rendered as blank. Fixes #99

### LINKED ISSUE #99: NA properties are sometimes rendered as "NA" in HTML
If an argument to a tag function is NA, then it should be turned into an empty HTML attribute:

```R
div(class = "abc", foo=NA)
#> <div class="abc" foo></div>
```

However, if there are two other arguments with the same name, then the NA will turn into `"NA"`:

```R
div(class = "abc", class = "xyz", foo=NA)
#> <div class="abc xyz" foo="NA"></div>
```

wch:


jjallaire:
I actually have a scenario right now where I depend on the `NA` behavior to add the `async` attribute to a `<script>` tag, e.g. `<script async src = "...">`. How can I do this after we merge this PR? How do we know that others aren't relying on this behavior?

PR #103, opened 2018-06-17
Improve error message in copyDependenciesToDir() when dependency file…

### LINKED ISSUE #102: copyDependenciesToDir() throws confusing error message when dependency file doesn't exist
```r
library(htmltools)

dep <- htmlDependency(
  name = "foo",
  version = "0.0.1",
  script = "foo.js",
  src = list(file = ""),
  package = "htmltools",
  all_files = FALSE
)

copyDependencyToDir(dep, "bar")
```

```r
Error in dirname(to) : a character vector argument expected
```

It'd be useful if the error message said something to the effect of: `Can't copy 'foo.js' since it doesn't exist`

cpsievert:
… does not exist, fixes #102

### DIFF
--- a/R/html_dependency.R
+++ b/R/html_dependency.R
@@ -305,6 +305,15 @@ copyDependencyToDir <- function(dependency, outputDir, mustWork = TRUE) {
   files <- if (dependency$all_files) list.files(dir) else {
     unlist(dependency[c('script', 'stylesheet', 'attachment')])
   }
+  if (any(!file.exists(files))) {
+    stop(
+      sprintf(
+        "Can't copy dependency files that don't exist: '%s'",
+        paste(files, collapse = "', '")
+      ),
+      call. = FALSE
+    )
+  }
   srcfiles <- file.path(dir, files)
   destfiles <- file.path(target_dir, files)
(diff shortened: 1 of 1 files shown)


R/html_dependency.R
```diff
@@ -305,6 +305,15 @@ copyDependencyToDir <- function(dependency, outputDir, mustWork = TRUE) {
   files <- if (dependency$all_files) list.files(dir) else {
     unlist(dependency[c('script', 'stylesheet', 'attachment')])
   }
+  if (any(!file.exists(files))) {
+    stop(
+      sprintf(
+        "Can't copy dependency files that don't exist: '%s'",
+        paste(files, collapse = "', '")
+      ),
+      call. = FALSE
```
wch: Is there a particular reason to use `call. = FALSE` here? I know that some people like to use that in general for error messages, but unless there's a specific reason here, I'd rather leave it off to be consistent with the rest of the code here.

cpsievert: > Is there a particular reason to use call. = FALSE here?

I think it reads slightly better, but I have no strong feelings, so I removed it

PR #105, opened 2018-06-26
Allow partial path in save_html #104

### LINKED ISSUE #104: save_html cannot work with relative path
Hi, I'm using the `save_html` function in one of my packages and a user discovered this bug: https://github.com/haozhu233/kableExtra/issues/220

It seems like we cannot use `save_html` with relative path. For example, if you do 
`save_html("aaa", "Desktop/aaa.html")`
it will generate an error.

haozhu233:
Fix #104

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -8,6 +8,8 @@ htmltools 0.3.6.9001
 
 * Fixed #99: `NA` attributes were sometimes rendered as `"NA"` in the HTML,
   instead of being blank. (#100)
+  
+* Fixed #104: `save_html()` cannot recognize relative path. (#105)
 
 
 htmltools 0.3.6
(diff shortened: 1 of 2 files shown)


haozhu233:
Any updates on this?

cpsievert:
@haozhu233 if you could also provide some minimal examples/tests that demonstrate the fix, we'd be more likely to take this on, thanks!

yihui (approved):
@cpsievert I feel this is a pretty bad bug and we should definitely fix it. This PR is relatively simple, and I'd probably just merge it (after adding a NEWS item). That said, it's easy to create a minimal example that demonstrates the bug, so I just made one:

```r
owd = setwd(tempdir())
dir.create('foo')
htmltools::save_html(DT::datatable(iris), 'foo/bar.html')
setwd(owd)
```

This will throw an error.

cpsievert:
Thank you for this nice contribution @haozhu233, and sorry for the delay! We'll take this over the finish line in #192

PR #108, opened 2018-08-23
Un-export knit_print methods

wch:
This PR un-exports the `knit_print` methods and registers them as S3 methods when both knitr and htmltools are loaded.

With R >=3.5.0, htmltools must be **attached** in order for the methods to work. There's an exception when shiny is attached: because shiny imports and re-exports these methods, they will also work if shiny is attached.

One consequence is that, because htmltools no longer exports these functions, Shiny can't import them. So Shiny will need to be updated to stop doing that.

Note that the CRAN version of Shiny (1.1.0) cannot be installed with this version of htmltools -- Shiny will fail to build because it can't import the `knit_print` methods from htmltools. If this version of htmltools is installed with an existing Shiny 1.1.0 installation, I believe that Shiny will no longer be able to be loaded.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,8 +1,7 @@
 Package: htmltools
 Type: Package
 Title: Tools for HTML
-Version: 0.3.6.9001
-Date: 2017-04-26
+Version: 0.3.6.9002
 Author: RStudio, Inc.
 Maintainer: Joe Cheng <joe@rstudio.com>
 Description: Tools for HTML generation and output.
--- a/NEWS
+++ b/NEWS
@@ -1,5 +1,9 @@
-htmltools 0.3.6.9001
+htmltools 0.3.6.9002
 --------------------------------------------------------------------------------
+
+* The `knit_print` methods are explicitly registered as S3 methods when both
+  knitr and htmltools are loaded. This means that neither shiny nor htmltools
+  need to be attached for the `knit_print` methods to work. (#108)
 
 * Updated RcppExports for new version of Rcpp. (#93)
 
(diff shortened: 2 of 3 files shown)


jcheng5:
> Note that the CRAN version of Shiny (1.1.0) cannot be installed with this version of htmltools -- Shiny will fail to build because it can't import the knit_print methods from htmltools. If this version of htmltools is installed with an existing Shiny 1.1.0 installation, I believe that Shiny will no longer be able to be loaded.

Will Shiny 1.2.0 be able to work with both older and newer htmltools? Or do we really need to do a completely synchronized release?

wch:
Here are the combinations:

* Old shiny, old htmltools: this is the current state of the world. `knit_print` methods are available if either shiny or htmltools is attached.
* New shiny, new htmltools: everything works -- `knit_print` methods work even if neither shiny nor htmltools are attached.
* New shiny, old htmltools: shiny works, but `knit_print` methods are available only if htmltools is attached. (Attaching shiny doesn't make them available.) Note that in the Shiny PR (rstudio/shiny#2166), I bumped the htmltools dependency so that this case shouldn't ever happen.
* Old shiny, new htmltools: Shiny will not build in this case. If it is already installed and htmltools upgraded, then shiny will fail to load.

So there is a problem with the upgrade path. We can't release htmltools first because it breaks shiny. If we released Shiny first (I could remove the htmltools version bump), then `knit_print` could be broken for users.

Maybe the best solution is to continue to export the `knit_print` methods for now, in addition to the registration; then we release new Shiny (which doesn't re-export them), then at some point in the future, after we can stop exporting them from htmltools.

wch:
I've changed this PR so that the only change is the the explicit registration of S3 methods; the `knit_print` methods are still exported. This avoids the shiny/htmltools upgrade issues described above.

PR #109, opened 2018-08-23
Better tag errors

wch:
Prior this change, the errors from tag functions when there are extra commas are pretty useless:

```
> div("a", "b", )
Error in tag("div", list(...)) : argument is missing, with no default

> tags$div("a", "b", )
Error in tag("div", list(...)) : argument is missing, with no default
```

After this change:

```
> div("a", "b",)
Error in div("a", "b",  ) : argument is missing, with no default

> tags$div("a", "b",)
Error in tags$div("a", "b",  ) : argument is missing, with no default
```

This PR also:

* Programmatically generates the tag functions instead of embedding a literal string in each one.
* Simplifies the directly-exported tag functions (like `div`) by removing the wrapper function.


wch:
The change that improves the error message is just that `list(...)` is evaluated before being passed to the `tag()` function. To illustrate:

```R
# Old version
f <- function(...) {
  tag("div", list(...))
}

f(1, 2, )
# Error in tag("div", list(...)) : argument is missing, with no default


# New version
g <- function(...) {
  contents <- list(...)
  tag("div", contents)
}

g(1, 2, )
# Error in g(1, 2, ) : argument is missing, with no default
```

PR #110, opened 2018-08-24
Un-export knit_print methods

wch:
This PR is a follow-up to #108. This one actually un-exports the `knit_print` methods. It should not be merged until, say, a year after the updated version of Shiny has been released.


schloerke:
It is 2021. @wch , Can this PR be merged?

PR #113, opened 2018-09-19
Bypassing dependency overwrite

Haunfelder:
This pull request adds the ability to skip overwriting html dependencies if they already exist in the function copyDependenciesToDir. Intended to be used with rmarkdown::render to speed up rendering by decreasing I/O time.

### DIFF
--- a/R/html_dependency.R
+++ b/R/html_dependency.R
@@ -252,6 +252,7 @@ urlEncodePath <- function(x) {
 #' @param dependency A single HTML dependency object.
 #' @param outputDir The directory in which a subdirectory should be created for
 #'   this dependency.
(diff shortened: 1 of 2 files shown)


R/html_dependency.R
```diff
@@ -290,6 +291,12 @@ copyDependencyToDir <- function(dependency, outputDir, mustWork = TRUE) {
   } else dependency$name
   target_dir <- file.path(outputDir, target_dir)
 
+  # if overwrite is false check to see if the file already exists
+  if(!overwrite_dir & dir_exists(target_dir)){
```
yihui: The condition `dir_exists(target_dir)` may be a little weak. A stronger version is to check if all filenames in the target dir and the dependency object are identical. This makes sure when the dependency is updated (in future versions of the package where the dependency is introduced), you'll get the new version.

PR #117, opened 2018-11-03
Open file connection with native encoding before writing.

hongyuanjia:
Hi!

Actually this comes when I am trying to convert ggplot plots using `plotly::ggplotly()`. It always issues a warning which comes from `htmltools:::save_html` complaining about invalid character and a blank web page is opened.

```r
getOption("encoding")
#> "UTF-8"

Warning message:
In writeLines(html, file, useBytes = TRUE)  :
    invalid char string in output conversion
```

If I set encoding from `"UTF-8"` to `"native.enc"`, the problem is gone.

As shown in this brilliant blog on [string encoding in R](http://kevinushey.github.io/blog/2018/02/21/string-encoding-and-r/), the safest way to write UTF-8 characters on Windows is manually open the connection with native encoding before writing.

Hope this PR will help. Thanks.

### DIFF
--- a/R/html_print.R
+++ b/R/html_print.R
@@ -98,9 +98,8 @@ save_html <- function(html, file, background = "white", libdir = "lib") {
             "</html>")
 
   # write it
-  file_nat <- file(file, open = "w+", encoding = "native.enc")
-  writeLines(enc2utf8(html), file_nat, useBytes = TRUE)
-  close(file_nat)
+  opts = options(encoding = "native.enc"); on.exit(options(opts), add = TRUE)
+  writeLines(enc2utf8(html), file, useBytes = TRUE)
 }
 
 
--- a/R/html_print.R
+++ b/R/html_print.R
@@ -98,7 +98,9 @@ save_html <- function(html, file, background = "white", libdir = "lib") {
             "</html>")
 
   # write it
-  writeLines(html, file, useBytes = TRUE)
+  file_nat <- file(file, open = "w+", encoding = "native.enc")
+  writeLines(enc2utf8(html), file_nat, useBytes = TRUE)
+  close(file_nat)
 }
 
 
--- a/R/html_print.R
+++ b/R/html_print.R
@@ -98,7 +98,8 @@ save_html <- function(html, file, background = "white", libdir = "lib") {
             "</html>")
 
   # write it
-  opts = options(encoding = "native.enc"); on.exit(options(opts), add = TRUE)
+  opts <- options(encoding = "native.enc")
+  on.exit(options(opts), add = TRUE)
   writeLines(enc2utf8(html), file, useBytes = TRUE)
 }
 


hongyuanjia:
@yihui I have no idea why checking failed on travis.

yihui:
Thanks!

cpsievert:
Hi @hongyuanjia, have you or would you be willing to sign RStudio's [individual](https://rstudioblog.files.wordpress.com/2017/05/rstudio_individual_contributor_agreement.pdf) or [corporate](https://rstudioblog.files.wordpress.com/2017/05/rstudio_corporate_contributor_agreement.pdf) contributor agreement? You can send the signed copy to jj@rstudio.com.

This change looks great to me overall. The only nit-pick I have is that you use `mode = "w+"` (reading and writing), but the default mode of `writeLines()` is write only (`mode = "wt"`). Moreover, I think I'd be in favor of approach that `xfun::write_utf8()` uses (setting/restoring the `encoding` option rather than manually opening/closing the file connection)

```r
opts = options(encoding = 'native.enc'); on.exit(options(opts), add = TRUE)
writeLines(enc2utf8(text), con, ..., useBytes = TRUE)
```

Lastly, just so I understand, do you only encounter this issue when you have `options(encoding = "UTF-8")` set? Here's a minimal reproducible example, but I'd be curious if there's another example that doesn't require setting the encoding to a non-native encoding

```r
Sys.getlocale()
#> 
options(encoding = "UTF-8")
library(htmltools)
tmp <- tempfile(fileext = ".html")
withr::with_dir(dirname(tmp), save_html(div("brûlée"), basename(tmp))
# brûlée is malformed in the resulting HTML file
browseURL(tmp)
```

hongyuanjia:
Hi @cpsievert. The signed contributor agreement has been sent to JJ and you. I have push a commit which uses option as suggested.

Below is the result of running the minimal example:

```r
Sys.getlocale()
# [1] "LC_COLLATE=English_United States.1252;LC_CTYPE=Chinese (Simplified)_China.936;LC_MONETARY=English_United States.1252;LC_NUMERIC=C;LC_TIME=English_United States.1252"
options(encoding = "UTF-8")
library(htmltools)
tmp <- tempfile(fileext = ".html")
withr::with_dir(dirname(tmp), save_html(div("brûlée"), basename(tmp)))
# brûlée appears as br没l茅e in the resulting HTML file
```

Currently, I did not encounter this issue with native encoding.

cpsievert:
Thank you @hongyuanjia. Your changes look good to me.

cpsievert:
Actually, could you change:

```r
opts = options(encoding = "native.enc"); on.exit(options(opts), add = TRUE)
```

to 

```r
opts <- options(encoding = "native.enc")
on.exit(options(opts), add = TRUE)
```

hongyuanjia:
Changed.

cpsievert:
Thanks for your contributions @hongyuanjia. We just merged #137, which seems to also fix this issue. Would you mind confirming that for us by installing from master?

```r
remotes::install_github("rstudio/htmltools")
```

hongyuanjia:
@cpsievert I just tested and I can confirm that current master branch has fixed this issue, with "UTF-8" encoding on Windows. Good to know so.

Close now.

PR #123, opened 2019-02-04
Don't use inline styles for html_print background color

jcheng5:


jcheng5:
@wch Do you want me to include this PR in the imminent htmltools release, or wait til the next one? I created it because it's relevant to bs4 theming.

PR #129, opened 2019-06-04
Add shim for system.file to work with devtools-loaded packages

wch:
This addresses rstudio/shiny#2468. Now when `devtools::test(, "bootstrap")` is called for shiny, it no longer has errors.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -18,6 +18,9 @@ htmltools 0.3.6.9003
 * The error message for trailing commas in tag functions now provides context
   and useful information. (#109)
 
+* Added a shim for `system.file()` so that htmltools works with `htmlDependency`
+  objects created by a package that was loaded with `devtools::load_all()`.
+  (#129)
 
 htmltools 0.3.6
 --------------------------------------------------------------------------------
--- a/R/shim.R
+++ b/R/shim.R
@@ -1,0 +1,62 @@
+# Borrowed from pkgload::dev_meta, with some modifications.
+devtools_loaded <- function(pkg) {
+  ns <- .getNamespace(pkg)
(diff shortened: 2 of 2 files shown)


cpsievert:
Hmm, when I run `devtools::test(, "bootstrap")`, I don't see any errors on https://github.com/rstudio/shiny/commit/d35c5c83201a218b22df11e778af6f8d8aade751

wch:
The reason to not use `pkgload::dev_meta()` is because then htmltools will take a dependency on `pkgload`.

wch:
> Just cause I'm curious, how does `find.package()` know to return the source path and not the installed path?

R doesn't have a concept of source path vs. installed path; it just has a concept of installed path. So when devtools loads a package, it tells R that the installed path is the source path.

cpsievert:
I was thinking we'd have pkgload in suggests (or not at all with a `getFromNamespace()` workaround), but I don't feel that strongly about it, so you have my 👍 either way

```r
if (system.file(package = "pkgload") != "") {
  return(!is.null(pkgload::dev_meta(package)))
} else if (is.null(ns) || is.null(ns$.__DEVTOOLS__)) {
  return(FALSE)
}
```

wch:
Thanks - I'd still rather not take any sort of dependency on `pkgload`. The worst that can happen if pkgload decides to change the `.__DEVTOOLS__` thing to something else is that this will no longer work, and the user will be in the same situation that exists currently.

Also, if we do use `pkgload::dev_meta`, then, if the user has pkgload installed, that line will cause the `pkgload` package to load (along with its dependencies) just to make this check, and that's way more heavyweight than is necessary here.

PR #131, opened 2019-06-11
Add argument to suppress whitespace

jcheng5:
- [x] Need to make `.noWS` a formal argument instead of an undocumented attribute
- [x] Need to document the enumerated values for .noWS
- [x] Validate .noWS arguments to ensure that nothing's misspelled
- [x] Update NEWS, bump version?
- [x] Why is .noWS whitespace-delimted? Shouldn't it just be a regular character vector?


R/tags.R
```diff
@@ -382,17 +383,27 @@ tagWrite <- function(tag, textWriter, indent=0, eol = "\n") {
...
+    noWS <- tag$attribs[[".noWS"]]
+    noWS <- strsplit(noWS, " +")[[1]]
+    tag$attribs[[".noWS"]] <- NULL
+  }
+  if ("before" %in% noWS || "outside" %in% noWS) {
+    textWriter$eatWS()
+  }
+
   # write tag name
-  textWriter(paste8(indentText, "<", tag$name, sep=""))
```
trestletech: Where does this indent get written now?

R/utils.R
```diff
@@ -0,0 +1,146 @@
...
+#
+# WSTextWriter relies on the caller distinguishing between writes of important
+# content, and writes of whitespace that may or may not be elided (`.$write()`
+# vs `.$writeWS()`).
+#
+# At any point, `eatWS` may be called, which will cause any recent `writeWS`
+# operations (i.e. those since either the beginning of time, or the most recent
+# `write` operation) to be undone, AND for any future `writeWS` calls to be
+# ignored. A call to `write` will be respected, and will restore normal
+# behavior.
```
trestletech: This API is a little confusing to me. Having to call `eatWS` after every `write` call seems a little tedious and error prone. I haven't looked at the usage enough, so maybe this ends up being simpler for the client, but this is surprising to me.

Would it be cleaner to have the client identify the portions of the text that it's writing out and then let the writer filter/scrub as it sees fit? i.e. 
```
writer$noWSMode(c("before", "inside")
writer$writeWS("\t", tag="before")
writer$resetWSMode()
```

then the writer can just filter out which WS it should be writing and which it shouldn't?

R/utils.R
```diff
@@ -0,0 +1,146 @@
...
+    initialize = function() {
+      private$writer <- TextWriter$new()
+    },
+    close = function() {
+      private$writer$close()
+    },
+    write = function(text) {
+      # If an error is going to happen while evaluating `text`, let's make it
+      # happen before we mutate any of our internal state
+      force(text)
```
trestletech: Note to self: figure out what this is.

R/utils.R
```diff
@@ -0,0 +1,146 @@
...
+
+      # Set a bookmark here, so that future calls to eatWS() don't affect this
+      # write
+      private$writer$savePosition()
+    },
+    # Write whitespace. If suppressWhitespace() was called and
+    # its effect has not been canceled, then this method no-ops.
+    # @param text Single element character vector containing only
+    #   whitespace characters
+    writeWS = function(text) {
```
trestletech: Should the "whitespace only" expectation be enforced? I realize why we would opt not to for performance, but I'm wondering if it's desirable.

R/utils.R
```diff
@@ -0,0 +1,146 @@
...
+    },
+    close = function() {
+      private$writer$close()
+    },
+    write = function(text) {
+      # If an error is going to happen while evaluating `text`, let's make it
+      # happen before we mutate any of our internal state
+      force(text)
+
+      # Stop eating whitespace
```
trestletech: Note to self: Nit: I think this comment is misleading. `setSuppress` blocks _all_ writes. We're doing this to restore writing functionality globally on the writer because this interface opts never to suppress a non-whitespace write.

trestletech: Do we really need this suppress behavior? Is it used by other callers? If not, it seems a little more intuitive IMHO to have this whitespace-suppressing writer manage whether or not it wants to write WS and then just omit the call to `write` in `writeWS` if it's in suppressing mode. That simplifies the interface on the textWriter, right?

R/utils.R
```diff
@@ -0,0 +1,146 @@
...
+  public = list(
+    initialize = function() {
+      # We use a file() here instead of textConnection() or paste/c to
+      # avoid the overhead of copying, which is huge for moderately
+      # large numbers of calls to .$write(). Generally when you want
+      # to incrementally build up a long string out of immutable ones,
+      # you want to use a mutable/growable string buffer of some kind;
+      # since R doesn't have something like that (that I know of),
+      # file() is the next best thing.
+      private$con <- file("", "w+b", encoding = "UTF-8")
```
trestletech: I can't tell from the `file` docs whether this is in-memory or on disk. It's an unlinked temporary file? I'm not sure what the behavior is there.

I ask because if this does live in memory I wonder if it would be worth having a configurable maximum buffer size.

R/utils.R
```diff
@@ -0,0 +1,146 @@
...
+      close(private$con)
+    },
+    # Write content; may or may not contain incidental whitespace,
+    # but whitespace that needs to be suppressible (like newlines
+    # and indents that the user did not explicitly request) should
+    # use writeWS() instead.
+    #
+    # Calling write() causes suppressWhitespace() to be cancelled.
+    #
+    # @param text Single element character vector
```
trestletech: Stale roxygen docs? This all feels specific to the whitespace writer, not the text writer.

R/utils.R
trestletech: I guess either way it's going to get read into memory during `readAll`, so maybe the `file` behavior is moot.

R/utils.R
```diff
@@ -0,0 +1,146 @@
...
+    # Mark the current writing position
+    savePosition = function() {
+      # Save the current write pos
+      private$marked <- seek(private$con, where = NA, rw = "write")
+    },
+    # Jump to the most recently marked position
+    restorePosition = function() {
+      if (!is.na(private$marked)) {
+        seek(private$con, private$marked, origin = "start", rw = "write")
+      }
```
trestletech: else error? warn?

tests/testthat/test-textwriter.r
```diff
@@ -0,0 +1,102 @@
...
+
+    tw$write("line one")
+    expect_identical(tw$readAll(), "line one")
+
+    tw$write("\nanother line")
+    expect_identical(tw$readAll(), "line one\nanother line")
+
+    tw$write("more content")
+    expect_identical(tw$readAll(), "line one\nanother linemore content")
+  })
```
trestletech: We should add tests for the UTF-8 behavior.

tests/testthat/test-textwriter.r
```diff
@@ -0,0 +1,102 @@
...
+  it("bookmarks are respected", {
+    tw <- TextWriter$new()
+    on.exit(tw$close())
+
+    tw$restorePosition()
+    expect_identical(tw$readAll(), "")
+
+    tw$write("foo")
+    tw$restorePosition()
+    expect_identical(tw$readAll(), "")
```
trestletech: I don't understand this one. Wouldn't we expect `foo`? `restorePosition` is a no op since it hasn't been set, so I'm not seeing why we'd get `""`?

trestletech: ohhh marked defaults to the beginning of the string. So we just undid our write.

R/tags.R
```diff
@@ -382,17 +383,27 @@ tagWrite <- function(tag, textWriter, indent=0, eol = "\n") {
 
   # compute indent text
   indentText <- paste(rep(" ", indent*2), collapse="")
+  textWriter$writeWS(indentText)
```
trestletech: Was this an intentional change to always write the indent regardless of whether or not tag is a character?

tests/testthat/test-textwriter.r
trestletech: See https://github.com/rstudio/shiny-examples/tree/master/022-unicode-chinese

R/utils.R
jcheng5: Maybe? I'm not even sure why there's a check there, it doesn't look to me like private$marked could be anything but a non-NA number? Maybe the original default for marked was NA, but then I changed it to 0.

trestletech:
I did some quick performance tests and it looks like this slows things down quite a bit. I haven't profiled to see where the time is going, but I'm seeing 60-100% extra time taken relative to master:

```
size <- 10000
system.time(l <- tagList(lapply(1:size, function(x){htmltools::tags$option(value=x, .noWS="outside before")})))
system.time(s <- as.character(l))

### .noWS branch
#> size <- 10000
#> system.time(l <- tagList(lapply(1:size, function(x){htmltools::tags$option(value=x)})))
#user  system elapsed 
#0.253   0.001   0.254 
#> system.time(s <- as.character(l))
#user  system elapsed 
#2.423   0.169   2.596

### .noWS branch with `.noWS="outside before"`
# > size <- 10000
# > system.time(l <- tagList(lapply(1:size, function(x){htmltools::tags$option(value=x, .noWS="outside before")})))
# user  system elapsed 
# 0.250   0.001   0.251 
# > system.time(s <- as.character(l))
# user  system elapsed 
# 2.774   0.179   2.957 

### master
#> size <- 10000
#> system.time(l <- tagList(lapply(1:size, function(x){htmltools::tags$option(value=x)})))
#user  system elapsed 
#0.246   0.005   0.252 
#> system.time(s <- as.character(l))
#user  system elapsed 
#1.574   0.014   1.592
```

trestletech:
After a bit of Travis thrashing followed by a force-push-fest to clean up the commit mess... ready for rereview, @jcheng5!

The one outstanding concern for me is the performance issue described in the comment above. I'm wondering if we should hold this PR back until we take on some of the C++ performance improvements. I think I could start on that work whether or not this is on master so long as we're not anticipating any serious conflicts to emerge in the coming couple of weeks.

trestletech:
I'm doing some experimentation on performance [here](https://github.com/rstudio/htmltools/commits/jeff/feature/stringbuilder). Raw performance notes from running this code:

```
size <- 1000
l <- tagList(lapply(1:size, function(x){htmltools::tags$option(value=x)}))
bench::mark(min_time=5,{
  s <- as.character(l)
})
```

By the end, it looks like we have an approach that will get us back to parity with master if we clean up that branch.

### master

```
# A tibble: 1 x 13
  expression                 min median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc total_time result memory time 
  <bch:expr>               <bch> <bch:>     <dbl> <bch:byt>    <dbl> <int> <dbl>   <bch:tm> <list> <list> <lis>
1 { s <- as.character(l) } 152ms  173ms      5.71     430KB     7.38    58    75      10.2s <chr … <df[,… <bch…
# … with 1 more variable: gc <list>
```

### binary connection-based

```
# A tibble: 1 x 13
  expression                 min median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc
  <bch:expr>               <bch> <bch:>     <dbl> <bch:byt>    <dbl> <int> <dbl>
1 { s <- as.character(l) } 213ms  239ms      4.20     508KB     8.80    21    44
# … with 5 more variables: total_time <bch:tm>, result <list
```

### buffer-based instead of connection (b09ee)

```
# A tibble: 1 x 13
  expression                 min median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc
  <bch:expr>               <bch> <bch:>     <dbl> <bch:byt>    <dbl> <int> <dbl>
1 { s <- as.character(l) } 295ms  326ms      3.05     129MB     11.3    16    59
# … with 5 more variables: total_time <bch:tm>, result <list>, memory <list>,
#   time <list>, gc <list>
```

### buffer-based with periodic string collection (345eff0)

```
# A tibble: 1 x 13
  expression                 min median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc
  <bch:expr>               <bch> <bch:>     <dbl> <bch:byt>    <dbl> <int> <dbl>
1 { s <- as.character(l) } 220ms  235ms      4.28    4.39MB     9.15    22    47
# … with 5 more variables: total_time <bch:tm>, result <list>, memory <list>,
#   time <list>, gc <list>
```

### buffer-based with string collection and double-assign to avoid copies

```
# A tibble: 1 x 13
  expression                 min median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc
  <bch:expr>               <bch> <bch:>     <dbl> <bch:byt>    <dbl> <int> <dbl>
1 { s <- as.character(l) } 200ms  221ms      4.51    40.1MB     9.42    23    48
# … with 5 more variables: total_time <bch:tm>, result <list>, memory <list>,
#   time <list>, gc <list>
```

### Same with portable=FALSE (aa8c3ad)

```
# A tibble: 1 x 13
  expression                 min median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc total_time
  <bch:expr>               <bch> <bch:>     <dbl> <bch:byt>    <dbl> <int> <dbl>   <bch:tm>
1 { s <- as.character(l) } 193ms  210ms      4.74     293KB     10.1    24    51      5.07s
# … with 4 more variables: result <list>, memory <list>, time <list>, gc <list>
```

### Same after eliminating TextConnection and just using WS directly (de642607)

```
# A tibble: 1 x 13
  expression                 min median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc total_time result memory time 
  <bch:expr>               <bch> <bch:>     <dbl> <bch:byt>    <dbl> <int> <dbl>   <bch:tm> <list> <list> <lis>
1 { s <- as.character(l) } 1.71s  1.77s     0.566    9.85MB     9.34     6    99      10.6s <chr … <df[,… <bch…
# … with 1 more variable: gc <list>

### after removing R6 and using a closure directly (5e5bab0c)

```
# A tibble: 1 x 13
  expression                 min median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc total_time result memory time 
  <bch:expr>               <bch> <bch:>     <dbl> <bch:byt>    <dbl> <int> <dbl>   <bch:tm> <list> <list> <lis>
1 { s <- as.character(l) } 142ms  166ms      5.74     271KB     7.72    58    78      10.1s <chr … <df[,… <bch…
# … with 1 more variable: gc <list>
```

File: R/tags.R [hunk 12]
```diff
@@ -382,17 +402,26 @@ tagWrite <- function(tag, textWriter, indent=0, eol = "\n") {
...
   if (is.character(tag)) {
-    textWriter(indentText)
-    textWriter(normalizeText(tag))
-    textWriter(eol)
+    textWriter$write(normalizeText(tag))
+    textWriter$writeWS(eol)
     return (NULL)
+  } else if (!is.null(tag$.noWS)){
+    .noWS <- tag$.noWS
+    validateNoWS(.noWS)
```
trestletech: Move that to tag definition time.

R/tags.R
```diff
@@ -382,17 +402,26 @@ tagWrite <- function(tag, textWriter, indent=0, eol = "\n") {
...
 
   # Check if it's just text (may either be plain-text or HTML)
   if (is.character(tag)) {
-    textWriter(indentText)
-    textWriter(normalizeText(tag))
-    textWriter(eol)
+    textWriter$write(normalizeText(tag))
+    textWriter$writeWS(eol)
     return (NULL)
+  } else if (!is.null(tag$.noWS)){
```
trestletech: Remove conditional, we don't need it.

R/tags.R
```diff
@@ -356,7 +366,8 @@ tag <- function(`_tag_name`, varArgs) {
   structure(
     list(name = `_tag_name`,
       attribs = attribs,
-      children = children),
+      children = children,
+      .noWS = .noWS),
```
trestletech: Try omitting the `.noWS` totally if NULL and see if that hashes in the test revert so we don't change everything.

amrrs:
@jcheng5 @wch could it be used with `htmlTemplate()`. It's still adding linebreaks

jcheng5:
@amrrs Oh, that's a good point. I think we'd actually need to add this feature to `HTML()` as well. This code:

```
div(HTML("a"), HTML("b"), .noWS = c("after-begin", "before-end"))
```

renders like:

```
<div>a
  b</div>
```

There's not currently a way AFAIK to remove the whitespace between `a` and `b`, and that's basically what we'd need to make `htmlTemplate` eat the whitespace.

jcubic:
What about space around `{{ ?? }}`?

jcheng5:
@jcubic That's what I was referring to in my previous comment, sorry it wasn't clear. `htmlTemplate` is essentially implemented by stringing together `HTML()` segments, as soon as we support whitespace eating around those segments then `{{ ... }}` can be fixed.

PR #132, opened 2019-07-10
Improve performance of writer.

trestletech:
- Use a buffer to cache writes instead of writing and seeking on a binary unliked connection. See https://wch.github.io/string_builder/ for impact on performance.
 - collapse the textwriter and wstextwriter into a single entity to spare all the extra dispatch
 - eliminate R6 and just use closures to spare ourselves the overhead
 - migrate from paste8 to paste0 since write UTF8-encodes all inputs.

TODO: 

 - [x] confirm that paste0 won't mess up any encoding

### on master (eea92)

```
> size <- 1000
> l <- tagList(lapply(1:size, function(x){htmltools::tags$option(value=x)}))
> bench::mark(min_time=10,{
+   s <- as.character(l)
+ })
# A tibble: 1 x 13
  expression                 min median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc
  <bch:expr>               <bch> <bch:>     <dbl> <bch:byt>    <dbl> <int> <dbl>
1 { s <- as.character(l) } 118ms  162ms      6.24     623KB     18.7    63   189
# … with 5 more variables: total_time <bch:tm>, result <list>, memory <list>,
#   time <list>, gc <list>
```

### on joe/feature/nows (b4dcc)

```
> size <- 1000
> l <- tagList(lapply(1:size, function(x){htmltools::tags$option(value=x)}))
> bench::mark(min_time=10,{
+   s <- as.character(l)
+ })
# A tibble: 1 x 13
  expression                 min median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc
  <bch:expr>               <bch> <bch:>     <dbl> <bch:byt>    <dbl> <int> <dbl>
1 { s <- as.character(l) } 219ms  257ms      3.89     499KB     7.08    39    71
# … with 5 more variables: total_time <bch:tm>, result <list>, memory <list>,
#   time <list>, gc <list>
```

### on jeff/feature/fast-nows(a038)

```
> size <- 1000
> l <- tagList(lapply(1:size, function(x){htmltools::tags$option(value=x)}))
> bench::mark(min_time=10,{
+   s <- as.character(l)
+ })
# A tibble: 1 x 13
  expression                 min median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc
  <bch:expr>               <bch> <bch:>     <dbl> <bch:byt>    <dbl> <int> <dbl>
1 { s <- as.character(l) } 130ms  163ms      5.88     895KB     16.7    59   167
# … with 5 more variables: total_time <bch:tm>, result <list>, memory <list>,
#   time <list>, gc <list>
```


R/utils.R
```diff
@@ -11,132 +9,113 @@
...
 
-      # Set a bookmark here, so that future calls to eatWS() don't affect this
-      # write
-      private$writer$savePosition()
+      # If our `marked` position has advanced well into the buffer,
+      # we can go ahead and collapse the string up to the marked position
+      # to free up room in the buffer.
+      if (marked >= length(buffer)/2) {
+        # Collapse the writes in the buffer up to the marked position
+        str <- paste(buffer[1:marked], collapse="")
```
wch: Better to use `seq_len(marked)` for safety. (I know that in this case, `marked` is definitely  >= 1, but it's still good practice).

R/utils.R
```diff
@@ -11,132 +9,113 @@
...
+        accumulated <<- paste0(accumulated, str)
+
+        # Move the remainder of the buffer up to the front and update the markers
+        remaining <- position - marked
+        if (remaining > 0){
+          buffer[1:(remaining)] <<-
```
wch: Same comment about `seq` as above.

R/utils.R
```diff
@@ -11,132 +9,113 @@
...
-      writeBin(raw, private$con)
-    },
-    # Return the contents of the TextWriter, as a single element
-    # character vector, from the beginning to the current writing
-    # position (normally this is the end of the connection, unless
-    # restorePosition() was called).
-    readAll = function() {
-      seek(private$con, where = 0, origin = "start", rw = "read")
-      s <- readChar(private$con, private$position, useBytes = TRUE)
       Encoding(s) <- "UTF-8"
```
wch: This shouldn't be necessary, since everything in `buffer` and `accumulated` should already be UTF-8, and `s` therefore should be as well. It was necessary when we were writing binary to a file and the reading it out, because the result of `readChar` would have been in the native encoding.

trestletech:
Incorporated feedback! I left the handling of multiple arguments in `write` as a TODO. We have some more complex interactions around `paste8` at the moment, but it seems like we are doubly encoding into utf8, so it might be wise to collapse these behaviors. It just didn't look like it was going to be free, but I may poke at it.

R/utils.R
```diff
@@ -11,132 +9,113 @@
...
+  position <- 0
+
+  # TRUE if we're eating whitespace right now, in which case calls to eatWS are no-ops.
+  suppressing <- FALSE
+
+  # Logic to do the actual write
+  writeImpl <- function(text) {
+    # If an error is going to happen while evaluating `text`, let's make it
+    # happen before we mutate any of our internal state
+    force(text)
```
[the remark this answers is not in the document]
trestletech: Experimenting here: https://github.com/rstudio/htmltools/compare/jeff/feature/fast-nows...jeff/feature/fast-nows2?expand=1

But so far not seeing a performance boost.

trestletech:
With the paste0 substitution in dce7d11, I'm now seeing a real performance boost over master.

```
> size <- 1000
> l <- tagList(lapply(1:size, function(x){htmltools::tags$option(value=x)}))
> bench::mark(min_time=10,{
+   s <- as.character(l)
+ })
# A tibble: 1 x 13
  expression                 min median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc
  <bch:expr>               <bch> <bch:>     <dbl> <bch:byt>    <dbl> <int> <dbl>
1 { s <- as.character(l) } 112ms  124ms      7.83     792KB     135.     4    69
# … with 5 more variables: total_time <bch:tm>, result <list>, memory <list>,
#   time <list>, gc <list>
```

But @wch pointed out that we need to confirm that the `paste0() + write()` combination still produces sane encoding.

R/utils.R
```diff
@@ -11,132 +9,112 @@
...
+  # The buffer into which we enter all the writes.
+  buffer <- character(bufferSize)
+
+  # The index storing the position in the buffer of the most recent write.
+  marked <- 0
+
+  # The index storing the position in the buffer of the most recent write or writeWS.
+  position <- 0
+
+  # TRUE if we're eating whitespace right now, in which case calls to eatWS are no-ops.
```
trestletech: eatWS -> writeWS

R/utils.R
```diff
@@ -11,132 +9,112 @@
...
+# Text is automatically converted to UTF-8 before being written.
+#' @param bufferSize The initial size of the buffer in which writes are stored.
+#'   The buffer will be periodically cleared, if possible, to cache the writes
+#'   as a string. If the buffer cannot be cleared (because of the need to be
+#'   able to backtrack to fulfill an `eatWS()` call), then the buffer size will
+#'   be doubled.
+#' @noRd
+WSTextWriter <- function(bufferSize=1024){
+  # A string into which we can collapse entries in the buffer if the buffer gets full.
+  accumulated <- ""
```
trestletech: Instead of having accumulated, try out having just collapsing into the first element in the buffer.

Make the first element in the array the collapsed writes that are persisted.
Make the second element in the array everything written since the last marked write position (all the whitespace).

trestletech:
Feedback incorporated

As of latest commit (e0262), we're about 10% faster than master.

```
# A tibble: 1 x 13
  expression                 min median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc total_time result
  <bch:expr>               <bch> <bch:>     <dbl> <bch:byt>    <dbl> <int> <dbl>   <bch:tm> <list>
1 { s <- as.character(l) } 112ms  145ms      6.97     796KB     7.57    70    76        10s <chr …
# … with 3 more variables: memory <list>, time <list>, gc <list>
```

trestletech:
Passing along some prior encoding research from Winston regarding whether or not we could further optimize `paste8`:

> Oh, wait, I just tried something else that has the bad behavior. Earlier, I said that `paste(x,y)` would convert both to UTF-8 before pasting, if one of them was UTF-8. But if you do `paste(c(x,y), collapse="")`, then it doesn't work that way.

```
Sys.setlocale(, "Chinese")
x <- "\377"
Encoding(x) <- "latin1"
x
#> [1] "ÿ" 

y <- "\U4E2d"  # Using \Uxxxx always is encoded as UTF-8
y
#> [1] "中"

xy <- paste0(x, y)
xy
#> [1] "ÿ中"
Encoding(xy)
#> [1] "UTF-8"


xy <- paste0(c(x, y), collapse = "")
xy
#> [1] "<ff>中"
Encoding(xy)
#> [1] "UTF-8"
```

jcheng5:
It somehow didn't sink in for me yesterday that you collapsed WSTextWriter and TextWriter into a single object. I would prefer that they be kept separate; TW presents a simple interface but has a tricky implementation, while WSTW presents a difficult-to-grok interface but has an extremely simple implementation. I found that splitting the respective logic into two separate classes made it easier to prove to myself that each one was individually correct, and that the composition of the two was also correct. With them combined into a single object now, it's harder for me to reason about.

Similarly, I think the logic in writeImpl that collapses the existing buffer should be made into a separate closure; again, I think having smaller parts that do one thing will be easier to reason about in isolation and in combination.

If you disagree let me know, I'm happy to chat about it.

Sorry for not catching this earlier, I'm sure you were expecting this to be ready to merge.

trestletech:
I agree that the buffer collapsing stuff should be pulled into its own closure. I'll do that!

I totally agree that the readability went way down when we collapsed them. The motivation was just performance to avoid the extra dispatch. I left the breadcrumbs over [here](https://github.com/rstudio/htmltools/pull/131#issuecomment-510141535) but we saw a 13.8% speedup when I collapsed the two writers into one. To me, getting the performance back to on-par with master was worth the readability hit, but if you disagree I'm happy to pull it back out.

R/utils.R
```diff
@@ -34,6 +34,31 @@ WSTextWriter <- function(bufferSize=1024){
   # TRUE if we're eating whitespace right now, in which case calls to writeWS are no-ops.
   suppressing <- FALSE
 
+  # Collapses the text in the buffer to create space for more writes. The first
+  # element in the buffer will be the concatenation of any writes up to the
+  # current marker. The second element in the buffer will be the concatenation
+  # of all writes after the marker.
+  collapseBuffer <- function(){
```
jcheng5: Can you add a space between `function()` and `{`? Sorry, that kind of stuff drives me crazy.

R/utils.R
```diff
@@ -34,6 +34,31 @@ WSTextWriter <- function(bufferSize=1024){
...
+    # Collapse the writes in the buffer up to the marked position into the first buffer entry
+    nonWS <- ""
+    if (marked > 0){
```
jcheng5: Can you add a space between `)` and `{`?

R/utils.R
```diff
@@ -34,6 +34,31 @@ WSTextWriter <- function(bufferSize=1024){
...
+      nonWS <- paste(buffer[seq_len(marked)], collapse="")
+    }
+
+    # Collapse any remaining whitespace
+    ws <- ""
+    remaining <- position - marked
+    if (remaining > 0){
```
jcheng5: Space between `)` and `{` here too

R/utils.R
trestletech: I smell an Rlint project for me to work on... ;)

jcheng5:
OK, if it's for performance reasons I'm fine with it. Thanks

PR #133, opened 2019-07-23
Add tests for different locales

trestletech:
(Please review each commit independently.)

Adds tests for different locales and demonstrates that Windows fails in some scenarios (see [AppVeyor results](https://ci.appveyor.com/project/trestletech/htmltools/builds/26189938) on 01fb94b) when using an abbreviated form of `paste8` that was being considered. The form proposed in 01fb94b does improve performance by ~20%, but we'd need additional checks to cover the scenarios that aren't reliably encoded which would make the implementation more fragile and -- perhaps -- not such a performance boost.

```
> test_check("htmltools")
-- 1. Failure: paste8 in Chinese locale works (@test-tags.r#749)  --------------
`xy` not identical to "<U+00FF>÷–".
1/1 mismatches
x[1]: "<ff>÷–"
y[1]: "<U+00FF>÷–"

-- 2. Failure: UTF-8 characters in templates (@test-template.R#54)  ------------
charToRaw(text) not identical to as.raw(c(195, 191)).
Lengths (4, 2) differ (comparison on first 2 components)
2 element mismatches

== testthat results  ===========================================================
OK: 225 SKIPPED: 0 WARNINGS: 0 FAILED: 2
1. Failure: paste8 in Chinese locale works (@test-tags.r#749) 
2. Failure: UTF-8 characters in templates (@test-template.R#54) 

Error: testthat unit tests failed
Execution halted
```

It appears that `paste` behaves differently when given multiple character arguments as parameters versus being given a character vector with elements of different encodings. In the former case, the strings are all converted to UTF-8 (at least if one of the elements is in UTF-8), but in the latter case the behavior isn't what we want.

These tests fail _only_ on Windows -- you'll notice that the Travis tests were happy all along.

### DIFF
--- a/tests/testthat/helper-locale.R
+++ b/tests/testthat/helper-locale.R
@@ -1,7 +1,7 @@
 is_locale_available <- function(loc){
   set_locale_failed <- FALSE
   tryCatch(
-    Sys.setlocale("LC_ALL", loc),
+    withr::with_locale(c(LC_COLLATE=loc), {}),
     warning = function(e){ set_locale_failed <<- TRUE }
   )
   !set_locale_failed
--- a/tests/testthat/helper-locale.R
+++ b/tests/testthat/helper-locale.R
@@ -1,0 +1,8 @@
+is_locale_available <- function(loc){
+  set_locale_failed <- FALSE
+  tryCatch(
+    Sys.setlocale("LC_ALL", loc),
+    warning = function(e){ set_locale_failed <<- TRUE }
+  )
+  !set_locale_failed
+}
--- a/tests/testthat/test-tags.r
+++ b/tests/testthat/test-tags.r
@@ -729,6 +729,30 @@ test_that("Latin1 and system encoding are converted to UTF-8", {
   )
 })
 
+test_that("paste8 in Chinese locale works", {
+  Sys.setlocale("LC_ALL", "Chinese")
+  x <- "\377"
+  Encoding(x) <- "latin1"
+  expect_identical(x, "ÿ")
+  expect_identical(Encoding(x), "latin1")
+
+  y <- "\U4E2d"  # Using \Uxxxx always is encoded as UTF-8
+  expect_identical(y, "中")
+  expect_identical(Encoding(y), "UTF-8")
+
+  xy <- paste8(x, y)
+  xy
+  expect_identical(xy, "ÿ 中")
+  expect_identical(Encoding(xy), "UTF-8")
+
+  xy <- paste8(c(x, y), collapse = "")
+  expect_identical(xy, "ÿ中")
+  expect_identical(Encoding(xy), "UTF-8")
+
+  # Restore default locale
+  Sys.setlocale("LC_ALL", "")
+})
+
 test_that("Printing tags works", {
   expect_identical(
     capture.output(print(tags$a(href = "#", "link"))),
--- a/tests/testthat/test-tags.r
+++ b/tests/testthat/test-tags.r
@@ -734,20 +734,20 @@ test_that("paste8 in Chinese locale works", {
   withr::with_locale(c(LC_COLLATE=loc, LC_CTYPE=loc, LC_MONETARY=loc, LC_TIME=loc), {
     x <- "\377"
     Encoding(x) <- "latin1"
-    expect_identical(x, "ÿ")
+    expect_identical(x, "\Uff")
     expect_identical(Encoding(x), "latin1")
 
     y <- "\U4E2d"  # Using \Uxxxx always is encoded as UTF-8
-    expect_identical(y, "中")
+    expect_identical(y, "\U4E2d")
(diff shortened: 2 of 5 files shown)


tests/testthat/test-template.R
```diff
@@ -29,21 +29,29 @@ test_that("Code blocks are evaluated and rendered correctly", {
...
-  Encoding(pat) <- "UTF-8"
-  expect_true(grepl(pat, html))
-
-  # If template is passed text_ argument, make sure it's converted from native
-  # to UTF-8.
-  latin1_str <- rawToChar(as.raw(0xFF))
-  Encoding(latin1_str) <- "latin1"
-  text <- as.character(htmlTemplate(text_ = latin1_str))
-  expect_identical(charToRaw(text), as.raw(c(0xc3, 0xbf)))
+  test_template <- function(locale=""){
```
trestletech: The implementation here is unchanged -- just adding the ability to set the locale prior to running it. When using the "alternate" `paste8` implementation proposed in 01fb9, this test fails only when running in the Chinese locale.

tests/testthat/test-template.R
```diff
@@ -29,21 +29,29 @@ test_that("Code blocks are evaluated and rendered correctly", {
...
+    latin1_str <- rawToChar(as.raw(0xFF))
+    Encoding(latin1_str) <- "latin1"
+    text <- as.character(htmlTemplate(text_ = latin1_str))
+    expect_identical(charToRaw(text), as.raw(c(0xc3, 0xbf)))
+
+    # Restore locale
+    Sys.setlocale("LC_ALL", "")
+  }
+  test_template("") # The default locale
+  test_template("Chinese") # Chinese locale
```
wch: Does this work on non-Windows platforms?

tests/testthat/test-tags.r
```diff
@@ -729,6 +729,30 @@ test_that("Latin1 and system encoding are converted to UTF-8", {
...
+  xy
+  expect_identical(xy, "ÿ 中")
+  expect_identical(Encoding(xy), "UTF-8")
+
+  xy <- paste8(c(x, y), collapse = "")
+  expect_identical(xy, "ÿ中")
+  expect_identical(Encoding(xy), "UTF-8")
+
+  # Restore default locale
+  Sys.setlocale("LC_ALL", "")
```
wch: I think it's safer to save the starting locale and then explictly restore it here (or better, with `on.exit()`), since it's possible that the user has set it to something other than the default.

tests/testthat/test-template.R
trestletech: It passes on my local machine -- but with a warning about Chinese not being available. So it's not a meaningful test.

tests/testthat/test-tags.r
trestletech: That was my first thought, too, but I couldn't get the semantics quite right so I gave it up:

```
> Sys.getlocale()
[1] "en_US.UTF-8/en_US.UTF-8/en_US.UTF-8/C/en_US.UTF-8/en_US.UTF-8"
> Sys.setlocale("en_US.UTF-8/en_US.UTF-8/en_US.UTF-8/C/en_US.UTF-8/en_US.UTF-8")
```

wch: Take a look at https://github.com/r-lib/withr/blob/master/R/locale.R to see how `withr::with_locale` works.

tests/testthat/test-template.R
wch: Maybe when you set the locale, do something like:

```R
set_locale_failed <- FALSE
tryCatch(
  Sys.setlocale(x),
  warning = function(e) set_locale_failed <<- TRUE
)

# Early return from test_template if we couldn't set the locale
if (set_locale_failed)
  return()

# do actual testing here
```

so that it it cleanly handles the warning and doesn't run irrelevant tests.

tests/testthat/helper-locale.R
```diff
@@ -0,0 +1,8 @@
+is_locale_available <- function(loc){
```
wch: If this function succeeds, it doesn't just test whether it's possible to set the locale -- it will also set the locale to the new one. See note below...

tests/testthat/test-tags.r
```diff
@@ -729,6 +729,31 @@ test_that("Latin1 and system encoding are converted to UTF-8", {
   )
 })
 
+test_that("paste8 in Chinese locale works", {
+  loc <- "Chinese"
+  testthat::skip_if_not(is_locale_available(loc), "Chinese locale not available")
```
wch: If `is_locale_available()` succeeds and returns TRUE, then the locale after this line will be `"Chinese"`. That means that after the `with_locale()` call below, the locale will still be `"Chinese"`.

tests/testthat/helper-locale.R
trestletech: Good catch! Fixed now, I believe.

PR #134, opened 2019-08-01
Add valid css units

atusy:
This PR adds valid css units such as `ch` and `rem`.

I'd also want to support `calc()`.
However, regex for fully validating it will be quite complex.
What about just add `calc(.*)` ?

### DIFF
--- a/R/tags.R
+++ b/R/tags.R
@@ -1438,8 +1438,9 @@ is.singleton <- function(x) {
 #'
 #' Single element character vectors must be \code{"auto"} or \code{"inherit"},
 #' or a number. If the number has a suffix, it must be valid: \code{px},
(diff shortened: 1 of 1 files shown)


jcheng5:
`calc(.*)` is totally fine IMHO, or I guess `^calc\\(.*\\)$`

Thanks for the contribution! Have you ever filed a [contributor agreement](https://github.com/rstudio/htmltools/blob/master/CONTRIBUTING.md) with RStudio before? If not, we'll need that on file before we can merge. Sorry for the inconvenience.

Thanks again!

atusy:
Thanks. I added `calc\\(.*\\)`

About the  contributor agreement, yes, I have when I contributed to `bookdown`.

PR #135, opened 2019-08-13
Add support for hanging commas

### LINKED ISSUE #128: Ignore trailing commas
``` r
library(htmltools)

div(
  p("one"),
  p("two"),
)
#> Error in tag("div", list(...)): argument is missing, with no default
```

<sup>Created on 2019-05-30 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1.9000)</sup>

The easiest way to do this would be to switch from `list()` to `rlang::list2()`

trestletech:
Closes #128 

Uses `rlang::dots_list` instead of `list` to support trailing commas with missing arguments a la

```
tagList("hi", )
div("one", )
```

I'll confess that I was guessing at which of the expectations in test-deps were going to turn out to be lists or dot_lists. That makes me wary that I might have missed something or that this may be an incomplete migration.

This does change the hash for singletons, but I added a test to confirm that the old hash will still work.

## Testing notes

Add a hanging comma at the end of some tags when defining a Shiny UI and it should work e.g.

```
tagList("hi", )
div("one", )
```

Multiple trailing commas should throw an error:

```
div("one", , )
```

Non-trailing commas should throw an error:

```
div( ,"one", )
```


NEWS
```diff
@@ -1,6 +1,8 @@
 htmltools 0.3.6.9003
 --------------------------------------------------------------------------------
 
+* Add support for trailing commas in tagLists and the predefined tags. (#128)
```
wch: The format is usually:

```
Fixed #128: .... (#135)
```

I'll just make the change directly.

hadley:
I’d suggest using `list2()` here, and I don’t think you should need to change the tests.

jcheng5:
(sorry just noticed this PR now)

This gives HTML tag functions tidyeval semantics now, doesn't it? Is there any situation where that could break existing code? (Maybe where an expression being passed to an HTML tag includes dplyr calls that need to interpret the !! themselves?)

Even if there are such cases, depending on how edgy they might be, it might not be a showstopper anyway. But the fear of such breaking changes was what stopped me from immediately making this change, hoping we might find or make a variant of list2 that didn't do tidyeval--or else convince ourselves that the edge cases are vanishingly edgy.

hadley:
Not general tidy evaluation semantics, just tidy dots semantics (i.e. you can use `!!!`)

jcheng5:
Thanks @hadley that's perfect

PR #145, opened 2019-10-17
Various improvements

cpsievert:
Inspired by work on bs4

### DIFF
--- a/R/tags.R
+++ b/R/tags.R
@@ -259,7 +259,7 @@ tagList <- function(...) {
 #' @rdname tag
 #' @export
 tagAppendAttributes <- function(tag, ...) {
-  tag$attribs <- c(tag$attribs, list(...))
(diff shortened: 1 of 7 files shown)


jcheng5:
@wch This should be ready for review. @cpsievert and I would love to get this merged in the next week or two if possible.

R/tags.R
```diff
@@ -299,14 +299,14 @@ tagAppendChild <- function(tag, child) {
 #' @rdname tag
 #' @export
 tagAppendChildren <- function(tag, ..., list = NULL) {
-  tag$children <- c(tag$children, c(list(...), list))
+  tag$children <- unname(c(tag$children, c(dots_list(...), list)))
```
wch: It would probably be good to have tests for the unnaming behavior, so that it doesn't accidentally creep back into the code in the future. Same for `tagSetChildren`.

PR #150, opened 2019-12-16
Add capturePlot/plotTag functions

jcheng5:
The `capturePlot` function takes a plot expression and saves it to a file. Think of this as `ggsave` but it also works for base graphics, and isn't as smart about figuring out what kind of file you're trying to create.

The `plotTag` function takes a plot expression and creates a self-contained `` tag.

Both functions let you control the graphics device that will be used to generate the data; by default, we use the algorithm in `shiny:::startPNG` with the addition of `ragg::agg_png` (preferred over Cairo but not over quartz) and minus respecting the `shiny.usecairo` option.

Both functions also let you control the arguments that are passed to the graphics device, via `...` and `deviceArgs`.

## Testing notes

There are some examples in  `?capturePlot` and `?plotTag`.


R/images.R
```diff
@@ -0,0 +1,204 @@
...
+#' \code{\link[Cairo:CairoPNG]{Cairo::CairoPNG()}} are used if their packages
+#' are installed. Otherwise, \code{\link[grDevices:png]{grDevices::png()}} is
+#' used.
+#'
+#' @return A graphics device function.
+#'
+#' @export
+defaultPngDevice <- function() {
+  if (capabilities("aqua")) {
+    grDevices::png
```
cpsievert: Just curious, why do you find quartz to be preferable?

R/images.R
```diff
@@ -0,0 +1,204 @@
...
+  } else if (nchar(system.file(package = "ragg"))) {
+    ragg::agg_png
+  } else if (nchar(system.file(package = "Cairo"))) {
```
cpsievert: I think you want either `nchar(system.file(package = "Cairo")) > 0` or `system.file(package = "Cairo") != ""` (I prefer the latter)

R/images.R
```diff
@@ -0,0 +1,204 @@
...
+  if (!is.function(device)) {
+    stop(call. = FALSE, "The `device` argument should be a function, e.g. `grDevices::png`")
+  }
+
+  expr <- rlang::enquo(expr)
+
+  tempFile <- missing(filename)
+
+  args <- rlang::list2(width = width, height = height, res = res)
+  argnms <- names(formals(device))
```
cpsievert: Maybe we should first check that `all(c("width", "height", "res") %in% argnms)`

R/images.R
```diff
@@ -0,0 +1,204 @@
...
+  if (!"..." %in% argnms) {
+    # Only include `width`, `height`, and `res` if the corresponding formal
+    # parameters are present.
+    args <- args[names(args) %in% argnms | names(args) == "filename"]
```
cpsievert: Why is this here? Should `args` contain `...` at this point?

R/images.R
```diff
@@ -0,0 +1,204 @@
...
+  }
+  args <- c(list(filename = filename), args, rlang::list2(...))
+
+  do.call(device, args)
+  dev <- grDevices::dev.cur()
+  on.exit(grDevices::dev.off(dev), add = TRUE, after = FALSE)
+
+  op <- graphics::par(mar = rep(0, 4))
```
cpsievert: Could we have a comment here as to why the margins are set in this way?

R/images.R
```diff
@@ -0,0 +1,204 @@
...
+  # Prevent examples() from prompting
+  grDevices::devAskNewPage(FALSE)
+  tryCatch(graphics::plot.new(), finally = graphics::par(op))
+
+  tryCatch({
+    result <- withVisible(rlang::eval_tidy(expr))
+    if (result$visible) {
+      capture.output(print(result))
```
cpsievert: Why is `capture.output()` used here?

R/images.R
```diff
@@ -0,0 +1,204 @@
...
+capturePlot <- function(expr, filename = tempfile(fileext = ".png"),
+  device = defaultPngDevice(), width = 400, height = 400, res = 72,
+  ...) {
+  if (!is.function(device)) {
+    stop(call. = FALSE, "The `device` argument should be a function, e.g. `grDevices::png`")
+  }
+
+  expr <- rlang::enquo(expr)
+
+  tempFile <- missing(filename)
```
cpsievert: This won't work if a `tempfile()` is supplied by the user. That said, I don't see a great way around this issue either. Maybe we could add a `cleanup` arg that could be used to explicitly request that we remove the `filename` if the expression fails?

R/images.R
```diff
@@ -0,0 +1,204 @@
...
+    result <- withVisible(rlang::eval_tidy(expr))
+    if (result$visible) {
+      capture.output(print(result))
+    }
+    filename
+  }, error = function(e) {
+    try({
+      if (tempFile && file.exists(filename))
+        unlink(filename)
+    })
```
cpsievert: Should this be `silent = TRUE`?

R/images.R
```diff
@@ -0,0 +1,204 @@
...
+#'   \code{1/96}.
+#' @param mimeType The MIME type associated with the \code{device}. Examples are
+#'   \code{image/png}, \code{image/tiff}, \code{image/svg+xml}.
+#' @param deviceArgs A list of additional arguments that should be included when
+#'   the \code{device} function is invoked.
+#' @param attribs A list of additional attributes that should be included on the
+#'   generated \code{<img>} (e.g. \code{id}, \code{class}).
+#' @param suppressSize By default, \code{plotTag} will include a \code{style}
+#'   attribute with \code{width} and \code{height} properties specified in
+#'   pixels. If you'd rather specify the image size using other methods (like
```
cpsievert: > By default, \code{plotTag} will include a \code{style} attribute with \code{width} and \code{height} properties specified in pixels.

Is this true? Looks to me as though there is no `style` attribute by default?

R/images.R
jcheng5: I don't know, I was copying this from shiny. @wch?

R/images.R
jcheng5: `svg` doesn't include `res`, once I saw that I decided all bets were off :joy:

R/images.R
jcheng5: If the graphics device function _doesn't_ have `...` in its signature, then there's a chance we can't safely pass it `width`, `height`, and/or `res`. For all three of those (OK, mostly `res`) we want to try to pass those values if the graphics device supports it.

`filename` is required though, as are any arguments the caller passed via `...`; if they passed a named or positional(!) argument via `...` that can't be passed to the graphics device, then that should be an error.

The reason we want different behavior for `width`, `height`, and `res` is because the operation could plausibly succeed without those, and we don't give the caller a way NOT to pass them if not for this logic here.

That being said, `names(args) == "filename"` is leftover from an earlier iteration, I'll remove that.

R/images.R
jcheng5: I have no idea--copied from Shiny. @wch?

R/images.R
jcheng5: `capture.output` is just there to prevent anything from going to the console, in case the return value is not actually a ggplot object or similar, just some random visible value.

I'll fix the `result$value` part, thanks!

R/images.R
jcheng5: Hmmm... or maybe unconditionally remove it? I mean, by passing a path to `capturePlot`, clearly you don't care about what's currently at that path...

jcheng5: Although equally, I don't think it's unreasonable that you called `capturePlot` with an explicit path, it failed, and a half-written file was left behind. The problem is that if `missing(filename)`, you don't even know what filename was used!

R/images.R
jcheng5: No, I wanted it to log to stderr but also for the original error to be the one that blows up the stack. (Anytime you're doing try/catch in a catch or finally, you're left with some pretty awkward choices IMHO...)

R/images.R
jcheng5:
```
> plotTag(plot(cars), alt = "foo")$attribs$style
[1] "width:400px;height:400px;"
```

R/images.R
jcheng5: Yeah... I think I like this as-is.

R/images.R
cpsievert: Ok, I was just curious

R/images.R
cpsievert: Ok, I was just curious

R/images.R
```diff
@@ -63,7 +63,7 @@ capturePlot <- function(expr, filename = tempfile(fileext = ".png"),
   if (!"..." %in% argnms) {
     # Only include `width`, `height`, and `res` if the corresponding formal
     # parameters are present.
-    args <- args[names(args) %in% argnms | names(args) == "filename"]
+    args <- args[names(args) %in% argnms]
   }
   args <- c(list(filename = filename), args, rlang::list2(...))
```
cpsievert: I don't think we can assume that most/all devices will have a filename argument (`svglite::svglite()` is one such example that uses `file` instead of `filename` so perhaps we should do some sort of partial matching on `file*`)

jcheng5:
svglite example:

```
plotTag(plot(pressure), "hi", device = svglite::svglite, pixelratio = 1/72, mimeType = "image/svg+xml")
```

DESCRIPTION
```diff
@@ -10,24 +10,29 @@ Depends:
...
+    base64enc,
     Rcpp,
     rlang
 Suggests:
     markdown,
     testthat,
-    withr
+    withr,
+    Cairo,
+    ragg,
```
wch: Extra trailing comma here.

R/images.R
wch: You could add an argument `cleanup = missing(filename)`. That way, it would have the same default behavior as it currently does, but the user could change it if they wanted to.

jcheng5:
Sorry @cpsievert last round

PR #155, opened 2020-02-18
Add parseCssColors() function

jcheng5:
`parseCssColors` takes CSS colors as strings, removes insignificant whitespace, decodes them to RGB (including `hsl()`), and returns strings in `#RRGGBB(AA)` form. (Note that [IE11 doesn't support #RRGGBBAA](https://caniuse.com/#feat=css-rrggbbaa); the point of this isn't to send the return values to the browser, but for them to be used in R.)

We've needed this in a few packages already (gt, shiny, bootstraplib); seems like a good time to robustify and centralize this logic.

## Testing notes

This feature has comprehensive unit tests.


R/colors.R
```diff
@@ -0,0 +1,315 @@
...
+  }
+
+  if (mustWork && any(!success)) {
+    stop("CSS color value(s) could not be parsed")
+  }
+
+  result
+}
+
+# Strips whitespace that isn't significant in the parsing of CSS colors.
```
cpsievert: What's an example of significant white-space (I don't immediately see why a naive approach to stripping white-space doesn't work)

R/colors.R
```diff
@@ -0,0 +1,315 @@
...
+# Takes a vector of strings whose elements represent a SINGLE hex color channel
+# (one or two hexadecimal digits) and return an integer in the range [0-255].
+decode_hex <- function(str) {
+  stopifnot(all(nchar(str) %in% c(1,2)))
+
+  # Single hex digits get doubled up
+  str <- ifelse(nchar(str) == 1, paste0(str, str), str)
+
+  res <- strtoi(str, 16)
+  stopifnot(!any(is.na(res)))
```
cpsievert: Could use `anyNA()` here (added in R3.1.0)?

R/colors.R
jcheng5: Here are some examples that *shouldn't* parse:

```
"# FF FF FF"
"rgb (64, 64, 64)"      # space between "rgb" and "(" isn't allowed
"hsl(180, 50 %, 50 %)"  # space before % isn't allowed
```

R/colors.R
jcheng5: TIL!

jcheng5:
Didn't realize the HSL-to-RGB conversion was in the spec: https://www.w3.org/TR/css-color-3/#hsl-color

R/colors.R
```diff
@@ -0,0 +1,315 @@
...
+      !!!strat$decoders
+    )
+    if (any(res$matching_rows)) {
+      result[!success][res$matching_rows] <- strat$encoder(res$values)
+      success[!success][res$matching_rows] <- TRUE
+    }
+  }
+
+  if (mustWork && any(!success)) {
+    stop("CSS color value(s) could not be parsed")
```
cpsievert:
```suggestion
stop(sprintf("CSS color value(s) could not be parsed: '%s'", paste0(str[!success], collapse = "', '")))
```

R/colors.R
```diff
@@ -0,0 +1,315 @@
...
+  as.integer(pmax(0, pmin(255, round(decode_float_identity(str) * 255))))
+}
+
+decode_float_identity <- function(str) {
+  stopifnot(all(grepl(paste0("^", regex_float, "$"), str)))
+  as.numeric(str)
+}
+
+encode_hex <- function(values) {
+  if (ncol(values) == 0 || nrow(values) == 0) return(character(0))
```
cpsievert: Is this supposed to cover the case where `match_and_decode()` has no matches (and thus returns no `values`)? In that case, `values` will be `NULL`, so perhaps this should be something like `if (!length(values)) return(character(0))`? 

Also, to make it more obvious what sort of input these encoding functions are expecting, perhaps you could also add `if (!is.matrix(values)) stop("Expected a matrix")` after the length check?

R/colors.R
```diff
@@ -0,0 +1,315 @@
...
+  if (ncol(values) > 4) {
+    stop("encode_hex called with too many columns")
+  }
+  if (!is.numeric(values) && !is.integer(values)) {
+    stop("encode_hex requires numeric or integer values")
+  }
+
+  if (is.numeric(values)) {
+    values <- round(values)
+  }
```
cpsievert:
```suggestion
if (is.numeric(values)) {
  values <- round(values)
} else {
  stop("encode_hex requires numeric values")
}
```

R/colors.R
```diff
@@ -0,0 +1,315 @@
...
+  colors <- ifelse(alpha == 0xFF,
+    sprintf("#%02X%02X%02X", red, green, blue),
+    sprintf("#%02X%02X%02X%02X", red, green, blue, alpha)
+  )
+  colors
+}
+
+encode_hsl <- function(values) {
+  if (ncol(values) == 0 || nrow(values) == 0) return(character(0))
```
cpsievert: Same as https://github.com/rstudio/htmltools/pull/155/files#r380982451

R/colors.R
```diff
@@ -0,0 +1,315 @@
...
+  }
+  rgb <- cbind(f(0), f(8), f(4)) * 255
+  rgb <- cbind(rgb, alpha)
+
+  encode_hex(rgb)
+}
+
+decode_color_keyword <- function(str) {
+  color <- css_color_keywords[str]
+  if (any(is.na(color))) {
```
cpsievert: Could use `anyNA()`?

tests/testthat/test-colors.R
```diff
@@ -0,0 +1,70 @@
+context("test-colors")
+
+colors_bad <- readLines(test_path("colors-bad.txt"))
+colors_good <- readLines(test_path("colors-bad.txt"))
```
cpsievert: typo in filename...also, these don't appear to be used in any tests?

R/colors.R
jcheng5: I went a slightly different way; my intention was to round only if `!is.integer()`.

R/colors.R
```diff
@@ -259,9 +264,24 @@ encode_hex <- function(values) {
   colors
 }
 
+# Convert HTML color keywords (plus "transparent") to integer matrix with 3
+# columns (r, g, b) and length(str) rows. Errors on invalid strings.
+decode_color_keyword <- function(str) {
+  color <- css_color_keywords[str]
+  if (anyNA(color)) {
+    stop("Invalid color keyword(s)")
```
cpsievert: I think you only want to do this if `mustWork = TRUE`?

jcheng5: No, the `decode_*` functions can be extremely strict about their input because they're protected by conservative regexes--in this case, `regex_color_keyword`.

PR #158, opened 2020-03-31
Remove rcpp

jimhester:
This removes Rcpp, instead using R's C API for the functionality. It then converts the C++ code to C (which was very minor).

I was not sure what variable naming scheme to use, as the existing code is a mix of camel case and snake case. I used snake case because the code I was adapting for this was snake case, but I can change the new code to use camel case if you prefer.


src/template.c
```diff
@@ -19,19 +28,26 @@ std::vector<std::string> template_dfa(CharacterVector x) {
...
-  if (x.length() != 1) {
-    stop("Input HTML must be a character vector of length 1");
+  if (Rf_xlength(x_sxp) != 1) {
+    Rf_error("Input HTML must be a character vector of length 1");
   }
-  std::string input = Rcpp::as<std::string>(x[0]);
-  std::vector<std::string> pieces(0);
+
+  SEXP str;
+  SEXP pieces = Rf_allocVector(STRSXP, 10);
```
jimhester: A used a initial allocation of 10, if you think that is too high we can lower it.

src/template.c
```diff
@@ -46,7 +62,9 @@ std::vector<std::string> template_dfa(CharacterVector x) {
       switch (c) {
       case '{':
         state = code;
-        pieces.push_back(input.substr(pieceStartIdx, i - pieceStartIdx - 1));
+        str = PROTECT(Rf_mkCharLenCE(input + pieceStartIdx, i - pieceStartIdx - 1, CE_UTF8));
```
jimhester: This assumes the input is UTF-8, if that is not a good assumption we can change this to just be `Rf_mkCharLen()`, which will use the native encoding

src/template.c
```diff
@@ -46,7 +62,9 @@ std::vector<std::string> template_dfa(CharacterVector x) {
...
+        REPROTECT(pieces = str_sxp_set(pieces, pieces_num++, str), pieces_idx);
```
jimhester: This `REPROTECT` code looks a bit weird, but it needed to be written this way to that rchk does not complain, we use very similar code in [glue](https://github.com/tidyverse/glue/blob/29eb0ff076810b5fbfac378072bff3ab77eba25b/src/glue.c#L149-L150)

src/template.c
```diff
@@ -153,11 +175,20 @@ std::vector<std::string> template_dfa(CharacterVector x) {
...
 
   // Add ending HTML piece
-  pieces.push_back(input.substr(pieceStartIdx, len - pieceStartIdx));
+  str = PROTECT(Rf_mkCharLenCE(input + pieceStartIdx, len - pieceStartIdx, CE_UTF8));
+  REPROTECT(pieces = str_sxp_set(pieces, pieces_num++, str), pieces_idx);
+  UNPROTECT(1);
+
+  if (pieces_num < Rf_xlength(pieces)) {
+    SETLENGTH(pieces, pieces_num);
+    SET_TRUELENGTH(pieces, pieces_num);
```
jimhester: These lines also seem weird, but setting the length like this does work and is the most efficient way to get a properly sized vector without an additional copy. This is what we do in [readr](https://github.com/tidyverse/readr/blob/d07eb692c412927fba47d319d8b66d9e64a82615/src/Collector.h#L48-L49)

jimhester:
I ran rchk on this locally and it was clean.

jimhester:
Also ran with valgrind and it looks good there as well

src/template.c
wch: 10 seems reasonable to me.  Does this need a `PROTECT`?

src/template.c
wch: The calling function does convert to UTF-8, but this function should probably throw a warning if the input is not UTF-8, just in case someone else wants to call it.

src/template.c
wch: Just curious - does the `UNPROTECT` affect this `REPROTECT`, or the previous `PROTECT`?

wch: Oh wait, I think I get it now.

src/template.c
wch: Can you add a comment about that?

src/template.c
wch: Can you add a comment with a quick explanation about this?

wch:
Looks good overall.

The `PROTECT` stuff makes me a bit nervous though -- with the tests you've already done, would they catch any possible PROTECT problems? I believe that R compiled with `--enable-strict-barrier` and `gctorture(TRUE)` are essentially an airtight test of that, but it's slow. (The r-debug docker image does have R compiled with strict barrier, though.)

jimhester:
I am pretty sure they would catch anything, but I can run with `gctorture()` to double check. I usually resort to `gctorture2(100)` or similar though since `gctorture()` is so slow...

jimhester:
After a relative eternity checking with `gctorture2(100)` and `--enable-strict-barrier` finished with a problem.

```
> library(testthat); pkgload::load_all(); library(htmltools); setwd("tests/testthat"); gctorture2(100); testthat::test_dir(".");
Loading htmltools
[1] 0
✔ |  OK F W S | Context
✔ |  60       | test-colors [1524.4 s]
✔ |  21       | dependencies [787.5 s]
✔ |   1       | images [127.6 s]
✔ |   4     1 | print [320.2 s]
────────────────────────────────────────────────────────────────────────────────
test-print.R:54: skip: Special characters are not re-encoded
Reason: On CRAN
────────────────────────────────────────────────────────────────────────────────
✔ | 152     1 | tags [1007.7 s]
────────────────────────────────────────────────────────────────────────────────
test-tags.r:807: skip: paste8 in Chinese locale works
Reason: Chinese locale not available is not TRUE
────────────────────────────────────────────────────────────────────────────────
✔ |  59     1 | templates [826.8 s]
────────────────────────────────────────────────────────────────────────────────
test-template.R:58: skip: UTF-8 characters in templates with Chinese locale
Reason: Chinese locale not available is not TRUE
────────────────────────────────────────────────────────────────────────────────
✔ |  27       | textwriter [265.3 s]
✔ |  10       | whitespace [79.1 s]

══ Results ═════════════════════════════════════════════════════════════════════
Duration: 5136.3 s

OK:       334
Failed:   0
Warnings: 0
Skipped:  3
```

wch:
With a problem? Also, wow, that’s slow.

jimhester:
Sorry, _without_ a problem

src/template.c
```diff
@@ -1,11 +1,20 @@
-#include <Rcpp.h>
-using namespace Rcpp;
+#include <Rinternals.h>
+
+SEXP str_sxp_set(SEXP x, int i, SEXP val) {
+  R_xlen_t len = Rf_xlength(x);
+  if (i >= len) {
+    len *= 2;
+    x = Rf_lengthgets(x, len);
```
kevinushey, line 7: Does this need to be protected? (Presumedly we have to use `REPROTECT()`)

src/template.c
jimhester: No, we protect it immediately below.

src/template.c
jimhester: The REPROTECT happens outside this function. I think this should be safe because `SET_STRING_ELT` won't allocate, and rchk doesn't have issue with this code. But it probably would be safer to pass the protection index to this function and just do the `REPROTECT` inside. I guess also currently we are REPROTECTing more often than we need to, I will change it.

jimhester: So actually what you suggest is how I had written this code originally (https://github.com/tidyverse/glue/commit/b0cfea86bc80856ef56cdae05532573ecbc35e47#diff-92efcf7db90903ecdc03a20503b14360) and unfortunately rchk has a false positive for this, it cannot track REPROTECTs across function boundaries. However converting the function to a macro solves the issue and also makes the code look more like a simple `SET_STRING_ELT()`, which is what I have done now.

src/template.c
jimhester: Detecting this is slightly more complicated that just checking the return of `Rf_getCharCE()` as you also have to check if the string is all ASCII (which have a CE_NATIVE encoding), but I have now added this warning.

src/template.c
jimhester: Done!

jimhester:
Ok I rewrote how the setting is being done, I think it now reads more clearly.

wch:
Great, can you add a NEWS item?

jimhester:
Ok, added a note to NEWS

Issue #84, opened 2017-04-26
Include tags$svg, tags$g, tags$path, etc.


jcheng5:
Also probably should check with latest relevant specs to make sure we're not missing any new tags.

cc @Stophface @timelyportfolio

wch:
Do we want to include the tags for elements inside the svg?

jcheng5:
Yep. I'll amend the title.

banbh:
Note that tags in htmltools do not convert "." in "-" (in contrast to `css()`). I.e., `css(a.b=1)` is `a-b:1;` but `p(a.b=1)` is `<p a.b="1"></p>`. This may be relevant because many SVG attributes have hyphens (see https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute) as opposed to HTML, where hardly any do.

Might it make sense for all tags to convert "." to "-" (in attributes)? Or perhaps to do so optionally, so that SVG tags could be made to do so?

Bogaso:
Is there any update on this request?

schloerke:
@banbh You are allowed to send in `-`'ed names.

```r
htmltools::p(`a-b` = 1)
#> <p a-b="1"></p>
htmltools::p("a-b" = 1)
#> <p a-b="1"></p>
```

PR #159, opened 2020-04-03
Autogenerate html/svg known_tags from mozilla website information

schloerke:
Fixes https://github.com/rstudio/htmltools/issues/84
Should fix https://github.com/alandipert/html2r/issues/4

Automatically pull 
* HTML elements from https://developer.mozilla.org/en-US/docs/Web/HTML/Element - all elements
* SVG elements from https://developer.mozilla.org/en-US/docs/Web/SVG/Element - only elements with documentation pages

Considered pulling information from these websites...
* W3 Schools
  * https://www.w3schools.com/tags/default.asp - Did not distinguish what should be considered obsolete
  * https://www.w3schools.com/graphics/svg_reference.asp - Had many extra tags such as `altGlyph`
* W3 Standard
  * https://www.w3.org/TR/2018/WD-html53-20181018/single-page.html - Could not distinguish what was not obsolete
  * https://svgwg.org/svg2-draft/single-page.html - Did not trust the resulting set that was web scraped


schloerke:
@wch 

> At the top of the script, add an explanation of what it does.

✅ 

> In tags.R, right above known_tags <- ...., add a comment saying how the content got there.

Moved the generated tags to a file and gave it a roxygen2-like comment.

Now `tags.R` includes `known_tags.R`.

schloerke:
@wch 

> Have you checked to make sure we haven't lost any tags? It's possible that we had a weird tag that's no longer there.

I pull in the tags from a CRAN install htmltools.  This set is merged with the new set.

I also double checked the obsolete tags on SVG and had accidentally included them.  They have been removed in https://github.com/rstudio/htmltools/pull/159/commits/0bc3a623cb7547bbf6f737c7aca9f5daa88a7e0e

PR #170, opened 2020-06-08
Don't use after = FALSE (only on R>=3.5 in capturePlot() and plotTage())

cpsievert:
This'll get the unit tests passing on R < 3.5 https://github.com/rstudio/htmltools/pull/168/checks?check_run_id=743874034#step:12:110


cpsievert:
This'll get the unit tests passing on R < 3.5 https://github.com/rstudio/htmltools/pull/168/checks?check_run_id=743874034#step:12:110

PR #157, opened 2020-02-19
should specify `perl=TRUE` in order to handle emoji strings on Windows

shrektan:
Closes #156 
Closes rstudio/DT#661

Issue #156, opened 2020-02-19
extractPreserveChunks() when string contains emoji unicode on Windows


shrektan:
When the string contains emoji Unicode like "\U0001F937" **on Windows** , `extractPreserveChunks()` returns unexpected results.

It leads to an issue that when the user compiles an Rmd file with such Unicode strings **on Windows**, the htmlwidgets of the HTML output file after the line contains the emoji string, can't be displayed properly, e.g., rstudio/DT#661

The cause of this issue I believe is that `gregexpr(perl = FALSE)`  (the default) can't handle such cases properly on Windows (I don't know the reason). By explicitly using `perl = TRUE` will fix this issue. 

https://github.com/rstudio/htmltools/blob/e07546ccb476a3f1c5cbe6178424635a886f8008/R/tags.R#L1170

# R code to reproduce

```r
x <- "<!--html_preserve-->chunck1<!--/html_preserve-->\U0001F937<!--html_preserve-->chunck2<!--/html_preserve-->"
htmltools::extractPreserveChunks(x)
```

# The output

```
$value
[1] "preserve054aaf381ec98343\U0001f937<preserve86252470130ed05d"

$chunks
preserve86252470130ed05d preserve054aaf381ec98343 
               "hunck2<"                "chunck1" 
```

# The session info

```
R version 3.6.2 (2019-12-12)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows 10 x64 (build 17763)

Matrix products: default

locale:
[1] LC_COLLATE=Chinese (Simplified)_China.936 
[2] LC_CTYPE=Chinese (Simplified)_China.936   
[3] LC_MONETARY=Chinese (Simplified)_China.936
[4] LC_NUMERIC=C                              
[5] LC_TIME=Chinese (Simplified)_China.936    

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] shiny_1.4.0

loaded via a namespace (and not attached):
 [1] Rcpp_1.0.3      rstudioapi_0.11 knitr_1.28      whisker_0.4    
 [5] magrittr_1.5    xtable_1.8-4    R6_2.4.1        rlang_0.4.2    
 [9] fastmap_1.0.1   tools_3.6.2     xfun_0.12       miniUI_0.1.1.1 
[13] clipr_0.7.0     htmltools_0.4.0 digest_0.6.23   processx_3.4.2 
[...]

PR #171, opened 2020-06-08
Use fixed = TRUE when finding preserve chunk locations

cpsievert:
Turns out the original fix for #156 (in #157) doesn't work in R 4.0.0 (and 4.0.1) due to a bug in R itself (which is now fixed, see https://github.com/rstudio/htmltools/pull/171#issuecomment-641469977)

https://github.com/rstudio/htmltools/runs/750362532#step:12:284

Tp workaround the problem, in this case, we can use `fixed = TRUE` instead of `perl = TRUE`

### DIFF
--- a/R/tags.R
+++ b/R/tags.R
@@ -1061,11 +1061,10 @@ extractPreserveChunks <- function(strval) {
   matches <- c(startmatches, endmatches)
   o <- order(matches)
   matches <- matches[o]
-  attr(matches, "match.length") <- c(
+  lengths <- c(
     attr(startmatches, "match.length", TRUE),
     attr(endmatches, "match.length", TRUE)
   )[o]
-  lengths <- attr(matches, "match.length", TRUE)
 
   # No markers? Just return.
   if (matches[[1]] == -1)
--- a/R/tags.R
+++ b/R/tags.R
@@ -1067,7 +1067,7 @@ extractPreserveChunks <- function(strval) {
   )[o]
 
   # No markers? Just return.
-  if (matches[[1]] == -1)
+  if (identical(unique(matches), -1))
     return(list(value = strval, chunks = character(0)))
 
   # If TRUE, it's a start; if FALSE, it's an end
--- a/R/tags.R
+++ b/R/tags.R
@@ -1056,7 +1056,15 @@ extractPreserveChunks <- function(strval) {
     strval <- paste(strval, collapse = "\n")
 
   # matches contains the index of all the start and end markers
-  matches <- gregexpr(pattern, strval, perl = TRUE)[[1]]
+  startmatches <- gregexpr(startmarker, strval, fixed = TRUE)[[1]]
+  endmatches <- gregexpr(endmarker, strval, fixed = TRUE)[[1]]
+  matches <- c(startmatches, endmatches)
+  o <- order(matches)
+  matches <- matches[o]
+  attr(matches, "match.length") <- c(
+    attr(startmatches, "match.length", TRUE),
+    attr(endmatches, "match.length", TRUE)
+  )[o]
   lengths <- attr(matches, "match.length", TRUE)
 
   # No markers? Just return.


cpsievert:
@shrektan would you be interested in reviewing this PR? Interestingly, with R >= 4.0 on Windows, `perl = TRUE` now returns different results with Unicode characters (which might be a bug in R itself?). A minimal example:

```r
regexpr("b", "foo\U0001F937bar", perl = TRUE)
```

on 3.6 this returns (as expected)

```r
[1] 5
attr(,"match.length")
[1] 3
```

but on R>=4.0, I see

```r
[1] 6
attr(,"match.length")
[1] 3
```

which leads to incorrect results in `extractPreserveChunks()` (cc @jcheng5)

R/tags.R
```diff
@@ -1056,7 +1056,15 @@ extractPreserveChunks <- function(strval) {
...
+  startmatches <- gregexpr(startmarker, strval, fixed = TRUE)[[1]]
+  endmatches <- gregexpr(endmarker, strval, fixed = TRUE)[[1]]
+  matches <- c(startmatches, endmatches)
+  o <- order(matches)
+  matches <- matches[o]
+  attr(matches, "match.length") <- c(
+    attr(startmatches, "match.length", TRUE),
+    attr(endmatches, "match.length", TRUE)
+  )[o]
   lengths <- attr(matches, "match.length", TRUE)
```
schloerke, line 1068:
```suggestion
  lengths <- c(
    attr(startmatches, "match.length", TRUE),
    attr(endmatches, "match.length", TRUE)
  )[o]
```

schloerke: The attr value isn't used and is only pulled directly out into `lengths` value

R/tags.R
```diff
@@ -1056,7 +1056,15 @@ extractPreserveChunks <- function(strval) {
     strval <- paste(strval, collapse = "\n")
 
   # matches contains the index of all the start and end markers
-  matches <- gregexpr(pattern, strval, perl = TRUE)[[1]]
+  startmatches <- gregexpr(startmarker, strval, fixed = TRUE)[[1]]
+  endmatches <- gregexpr(endmarker, strval, fixed = TRUE)[[1]]
+  matches <- c(startmatches, endmatches)
```
schloerke, line 1061: With this matching both start and end directly, if either start or end does not have a match, then one of the `matches` values will be `-1`.

Before this change, BOTH start and end had to be missing to fail the `No markers? Just return.` check.  Now it is `SOME` marker is missing, so just return.

Maybe we should check if both `startmatches` and `endmatches` ? I don't know of the right behavior at first thought

wch:
For future reference, the failing test is due to a bug introduced in R 4.0.
https://stat.ethz.ch/pipermail/r-devel/2020-June/079654.html

And it's fixed in R-devel: https://github.com/wch/r-source/commit/0a9dbc79cbf8ce86ae25f0e67642dfc0b64d9ae1

R/tags.R
cpsievert: FWIW, I had considered this, and decided not too in case, at a latter point, someone wants to use `attr(matches, "match.length")`, but I suppose it's not worth the effort

R/tags.R
cpsievert: Doesn't the early return here catch the case where both are missing?

https://github.com/rstudio/htmltools/pull/171/files#diff-72abdc1c2c5a6d65974d57ef9bd3666cR1070

schloerke: Yes. Were both are missing.

But if start is defined and end is missing, the ordered `matches[[1]]` will return -1 as the end matches was missing.

cpsievert: Ahh, right, addressed by  https://github.com/rstudio/htmltools/pull/171/commits/3a5457dd4e4324f80cc75f592ff8aae0a4043015

PR #179, opened 2020-09-21
Introducing htmlDependencyFunction()

cpsievert:
## TODO: 

- [ ] unit tests


R/tags.R
```diff
@@ -56,16 +56,6 @@ registerMethods <- function(methods) {
...
 
-depListToNamedDepList <- function(dependencies) {
-  if (inherits(dependencies, "html_dependency"))
-    dependencies <- list(dependencies)
-
-  if (is.null(names(dependencies))) {
-    names(dependencies) <- sapply(dependencies, `[[`, "name")
-  }
-  return(dependencies)
-}
```
cpsievert, line 67: Pretty sure this isn't being used anywhere 🤷‍♂️

jcheng5:
What's this for? I mean, I can see it's for delaying the evaluation of a dependency but when exactly do you need to delay it until? I'm wondering if you can use the already-existing `as.tags` S3 method for this?

cpsievert:
Thanks for the idea @jcheng5, closing as #180 should hopefully handle this case (lazy dependencies) as well as lazy tags

PR #180, opened 2020-09-22
Introducing tagFunction() for 'lazily' rendered tags (and dependencies)

cpsievert:
Adds `tagFunction()`, which can be used to generate tags and/or dependencies conditional on the rendering context:

``` r
library(htmltools)
myDivDep <- tagFunction(function() {
  if (isTRUE(getOption("useDep", TRUE))) {
    htmlDependency(
      name = "lazy-dependency",
      version = "1.0", src = ""
    )
  }
})
myDiv <- attachDependencies(div(), myDivDep)
renderTags(myDiv)
#> $head
#> 
#> 
#> $singletons
#> character(0)
#> 
#> $dependencies
#> $dependencies[[1]]
#> List of 10
#>  $ name      : chr "lazy-dependency"
#>  $ version   : chr "1.0"
#>  $ src       :List of 1
#>   ..$ file: chr ""
#>  $ meta      : NULL
#>  $ script    : NULL
#>  $ stylesheet: NULL
#>  $ head      : NULL
#>  $ attachment: NULL
#>  $ package   : NULL
#>  $ all_files : logi TRUE
#>  - attr(*, "class")= chr "html_dependency"
#> 
#> 
#> $html
#> <div></div>
withr::with_options(list(useDep = FALSE), renderTags(myDiv))
#> $head
#> 
#> 
#> $singletons
#> character(0)
#> 
#> $dependencies
#> list()
#> 
#> $html
#> <div></div>
```

<sup>Created on 2020-09-24 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>


R/tags.R
```diff
@@ -879,11 +887,14 @@ flattenTags <- function(x) {
...
     }
 
-  } else if (is.character(x)){
+  } else if (is.character(x)) {
     # This will preserve attributes if x is a character with attribute,
     # like what HTML() produces
     list(x)
 
+  } else if (is_html_dependency(x)) {
+    list()
```
cpsievert, line 896: Needed to prevent infinite recursion now that `as.tags.html_dependency()` returns a `tagList()` of length 1

R/tags.R
```diff
@@ -857,10 +865,13 @@ withTags <- function(code) {
...
   rewriteTags(x, function(uiObj) {
-    if (isTag(uiObj) || isTagList(uiObj) || is.character(uiObj))
+    if (isTag(uiObj) || isTagList(uiObj) || is.character(uiObj)) {
       return(uiObj)
-    else
-      return(tagify(as.tags(uiObj)))
+    }
+    if (is_html_dependency(uiObj)) {
+      return(as.tags(uiObj))
+    }
```
cpsievert, line 873: Needed to prevent infinite recursion now that as.tags.html_dependency() returns a tagList() of length 1

R/tags.R
```diff
@@ -256,6 +256,14 @@ tagList <- function(...) {
   return(lst)
 }
 
+#' @export
+tagFunction <- function(func) {
+  if (!is.function(func) || length(formals(func)) != 0) {
+    stop("`func` must be a functions with no formal arguments")
```
wch, line 262:
```suggestion
    stop("`func` must be a function with no arguments")
```

R/html_dependency.R
```diff
@@ -154,32 +154,54 @@ validateScalarName <- function(x, name = deparse(substitute(x))) {
 #'
 #' @export
 htmlDependencies <- function(x) {
-  attr(x, "html_dependencies", TRUE)
+  unname(c(
```
wch, line 157: This is kind of hard to understand when all the logic is nested like this. Can you change it have multiple statements (and use variables)?

R/tags.R
```diff
@@ -256,6 +256,14 @@ tagList <- function(...) {
   return(lst)
 }
 
+#' @export
+tagFunction <- function(func) {
+  if (!is.function(func) || length(formals(func)) != 0) {
+    stop("`func` must be a function with no arguments")
+  }
+  structure(func, class = "shiny.tag.function")
```
cpsievert, line 267:
```suggestion
  structure(func, class = c("shiny.tag.function", "function"))
```

wch: Why? Even without the explicit `"function"` class, `is.function(func)` will still be TRUE.

R/tags.R
```diff
@@ -679,17 +690,21 @@ takeHeads <- function(ui) {
 #'
 #' @param tags A tag-like object to search for dependencies.
 #' @param tagify Whether to tagify the input before searching for dependencies.
+#' @inheritParams resolveDependencies
 #'
 #' @return A list of \code{\link{htmlDependency}} objects.
 #'
 #' @export
-findDependencies <- function(tags, tagify = TRUE) {
+findDependencies <- function(tags, tagify = TRUE, resolveTagFunction = TRUE) {
```
wch, line 698: When would `resolveTagFunction` be FALSE? Would it make sense to export `resolveFunctionalDependencies` instead?

R/tags.R
```diff
@@ -698,7 +713,35 @@ findDependencies <- function(tags, tagify = TRUE) {
     }
   }
   childDeps <- unlist(lapply(children, findDependencies, tagify = FALSE), recursive = FALSE)
-  c(childDeps, if (!is.null(dep)) dep)
+  c(childDeps, if (!is.null(deps)) deps)
```
wch, line 716: The `!is.null` check can actually be removed. Because when an `if` statement isn't true, it the whole thing evaluates to `NULL`.

```
print(if (F) "hello")
#> NULL
```

R/tags.R
```diff
@@ -679,17 +690,21 @@ takeHeads <- function(ui) {
...
-findDependencies <- function(tags, tagify = TRUE) {
+findDependencies <- function(tags, tagify = TRUE, resolveTagFunction = TRUE) {
   if (isTRUE(tagify)) {
     tags <- tagify(tags)
   }
-  dep <- htmlDependencies(tags)
-  if (!is.null(dep) && inherits(dep, "html_dependency"))
-    dep <- list(dep)
+  deps <- htmlDependencies(tags)
+  if (!is.null(deps) && inherits(deps, "html_dependency"))
```
wch, line 703: The `is.null()` check is redundant, and can be removed.

R/tags.R
```diff
@@ -698,7 +713,35 @@ findDependencies <- function(tags, tagify = TRUE) {
...
+    if (is_html_dependency(x)) {
+      return(list(x))
+    }
+    if (!inherits(x, "shiny.tag.function")) {
+      stop(call. = FALSE, "HTML dependencies must be a list containing either ",
+           "htmlDependency()s or tagFunction()s")
+    }
+    y <- x()
+    if (is_html_dependency(y)) {
+      return(list(deps))
```
wch, line 733: `list(y)`?

R/tags.R
```diff
@@ -698,7 +713,35 @@ findDependencies <- function(tags, tagify = TRUE) {
...
+    }
+    if (!all(vapply(y, is_html_dependency, logical(1)))) {
```
wch, line 735: It's not entirely clear to me: if `y` is a non-list, will it enter this block and give the error message? Or will it throw some sort of (confusing) error?

R/tags.R
```diff
@@ -698,7 +713,35 @@ findDependencies <- function(tags, tagify = TRUE) {
...
+  c(childDeps, if (!is.null(deps)) deps)
+}
+
+resolveFunctionalDependencies <- function(deps) {
+  if (!length(deps)) {
+    return(deps)
+  }
+  deps <- lapply(deps, function(x) {
+    if (is_html_dependency(x)) {
+      return(list(x))
```
wch, line 725: Why is this wrapped in a list? Same for below, where it returns `list(deps)`.

R/tags.R
```diff
@@ -928,6 +971,17 @@ as.tags.shiny.tag.list <- function(x, ...) {
   x
 }
 
+#' @export
+as.tags.shiny.tag.function <- function(x, ...) {
+  y <- x()
+  # as.tags() doesn't currently have a method for a list of dependencies
+  if (is.list(y) && all(vapply(y, inherits, logical(1), "html_dependency"))) {
+    lapply(y, function(z) as.tags(z))
```
wch, line 979: Is the `function(z) as.tags(z)` necessary because of S3 dispatch issues?

R/tags.R
```diff
@@ -928,6 +971,17 @@ as.tags.shiny.tag.list <- function(x, ...) {
   x
 }
 
+#' @export
+as.tags.shiny.tag.function <- function(x, ...) {
+  y <- x()
+  # as.tags() doesn't currently have a method for a list of dependencies
+  if (is.list(y) && all(vapply(y, inherits, logical(1), "html_dependency"))) {
```
wch, line 978: Will a case like the following be handled properly:

```R
div(list(
  div(),
  htmlDependency("abc", "1.0", "")
))
```

That is, where it's a plain list that has some, but not all, html dependencies.

R/tags.R
```diff
@@ -76,13 +76,16 @@ depListToNamedDepList <- function(dependencies) {
...
 #'   path via \code{\link{system.file}} when the \code{package} attribute is
 #'   present in a dependency object.
+#' @param resolveTagFunction whether or not to resolve [tagFunction()] dependencies.
 #' @return dependencies A list of \code{\link{htmlDependency}} objects with
 #'   redundancies removed.
 #'
 #' @export
-resolveDependencies <- function(dependencies, resolvePackageDir = TRUE) {
-  # Remove nulls
-  deps <- dependencies[!sapply(dependencies, is.null)]
```
wch, line 85: Good to get rid of this. I really dislike `sapply`.

R/tags.R
cpsievert: @schloerke recommended it

R/tags.R
cpsievert: 🤦

R/tags.R
cpsievert: Yep

R/tags.R
cpsievert: Since it's possible for `x()` to return a list of dependencies, I think it's easiest to just always return a list of dependencies so we can do `unlist(deps, recursive = FALSE)`

R/tags.R
cpsievert: If `y` is zero-length it won't enter this block (but the `unlist()` should remove any zero-length return value, which is by design)

If `y` is non-list with positive length, it will enter this block and give the error message.

Now that I think of it, I think we need to handle the `isTag(y)` and `isTagList(y)` cases as well...

R/tags.R
wch: I think we should not have it -- typically if you add a class, you don't also add the base class. I can't think of any cases where a classed list or atomic vector has the base class also explicitly added.

R/tags.R
wch: Makes sense.

R/tags.R
wch: Do you mind adding a comment above, so that we'll remember this in the future? Otherwise someone may try to simplify it and break things.

wch:
We'll also need to have some extensive tests, to make sure we're not missing any edge cases.

R/tags.R
cpsievert: Good point, until we find some reason to not to resolve, I'm on-board with always resolving and exporting `resolveFunctionalDependencies()`

R/tags.R
cpsievert: I'm not sure I understand the question. Are you concerned about this sort of example? This sort of thing should generally work because `tagify()` calls itself recursively on the return value of `as.tags()` (i.e., it's ok if `as.tags()` returns 'non-tag' objects because `as.tags()` will still get called on those contents)

```r
str(as.tags(tagFunction(function() {
  div(list(
    div(),
    htmlDependency("abc", "1.0", "")
  ))
})))
```

```r
List of 3
 $ name    : chr "div"
 $ attribs : Named list()
 $ children:List of 1
  ..$ :List of 2
  .. ..$ :List of 3
  .. .. ..$ name    : chr "div"
  .. .. ..$ attribs : Named list()
  .. .. ..$ children: list()
  .. .. ..- attr(*, "class")= chr "shiny.tag"
  .. ..$ :List of 10
  .. .. ..$ name      : chr "abc"
  .. .. ..$ version   : chr "1.0"
  .. .. ..$ src       :List of 1
  .. .. .. ..$ file: chr ""
  .. .. ..$ meta      : NULL
  .. .. ..$ script    : NULL
  .. .. ..$ stylesheet: NULL
  .. .. ..$ head      : NULL
  .. .. ..$ attachment: NULL
  .. .. ..$ package   : NULL
  .. .. ..$ all_files : logi TRUE
  .. .. ..- attr(*, "class")= chr "html_dependency"
 - attr(*, "class")= chr "shiny.tag"
```

R/tags.R
```diff
@@ -928,6 +988,18 @@ as.tags.shiny.tag.list <- function(x, ...) {
...
+as.tags.shiny.tag.function <- function(x, ...) {
+  y <- x()
+  # as.tags() doesn't currently have a method for a list of dependencies
+  if (is.list(y) && all(vapply(y, is_html_dependency, logical(1)))) {
+    # In order for S3 dispatch to work properly, this can't simply be `lapply(y, as.tags)`
+    lapply(y, function(z) as.tags(z))
+  } else {
+    as.tags(y)
+  }
+}
```
schloerke, line 1001: While this isn't technically equivalent, I don't see the reason for _not_ calling `as.tags` for each element.
```suggestion
#' @export
as.tags.list <- function(x, ...) {
  tagList(lapply(x, function(z) {
    # only recurse if it contains an html_dependency
    if (is_html_dependency(z)) {
      return(as.tags(z))
    }
    # maintain backwards compatibility by NOT calling `as.tags(z)
    z
  }))
}
#' @export
as.tags.shiny.tag.function <- function(x, ...) {
  # evaluate and convert to tags
  as.tags(x())
}
```

schloerke (changes requested):

R/tags.R
cpsievert: I like the idea, but I'm worried about what this means for the `if (is.list(x) && !isTagList(x))` case...

schloerke: There already is a method called `as.tags.shiny.tag.list(x)` which is the class that `isTagList(x)` checks for.  

So I believe this case is covered already and that historical code path was a dead code path.

cpsievert: Ok, yea, but that code path isn't dead. In fact, if you change `unclass(x)` to `x` then you get:

```r
> tagify(structure(list(), class = "foo"))
Error: evaluation nested too deeply: infinite recursion / options(expressions=)?
```

So, _I think_ that line is there simply as a way out of the crazy recursion logic in `tagify()`...it seems like it'd be better to avoid the recursion in that case by just wrapping the list in a `tagList()` like you've suggested here.

BTW, it also seems like we can get away without doing the `lapply()` you've suggested here, thanks to the recursive nature of `tagify()`

cpsievert: Handled in 03d79ec and 740c3a3

schloerke (approved):

R/tags.R
```diff
@@ -81,12 +81,11 @@ depListToNamedDepList <- function(dependencies) {
...
 #' @export
 resolveDependencies <- function(dependencies, resolvePackageDir = TRUE) {
-  # Remove nulls
-  deps <- dependencies[!sapply(dependencies, is.null)]
+  deps <- resolveFunctionalDependencies(dependencies)
 
   # Get names and numeric versions in vector/list form
-  depnames <- sapply(deps, `[[`, "name")
-  depvers <- numeric_version(sapply(deps, `[[`, "version"))
+  depnames <- vapply(deps, function(x) x$name, character(1))
```
wch, line 87: Doesn't really matter here, but for future reference, it's slightly faster to do it the previous way, because there's one less function call for each element:

```R
vapply(deps, `[[`, "name", FUN.VALUE = character(1))
```

R/tags.R
```diff
@@ -928,6 +997,16 @@ as.tags.shiny.tag.list <- function(x, ...) {
...
 }
 
+#' @export
+as.tags.shiny.tag.function <- function(x, ...) {
+  as.tags(x())
+}
+
+#' @export
+as.tags.list <- function(x, ...) {
+  tagList(x)
```
wch, line 1007: This will wrap the list into a tagList, adding one level to the tree. I think it makes more sense to just convert it to a `tagList`.

R/tags.R
```diff
@@ -292,24 +327,35 @@ tagGetAttribute <- function(tag, attr) {
...
 #' @rdname tag
 #' @export
 tagSetChildren <- function(tag, ..., list = NULL) {
+  tag <- assert_shiny_tag(tag)
   tag$children <- unname(c(dots_list(...), list))
   tag
 }
 
+assert_shiny_tag <- function(tag) {
+  tag <- as.tags(tag)
```
wch, line 352: A few issues/questions:

* Why this line is necessary? It would help to have a comment explaining why this is here.
* If the function is going to modify the object and return it, a name with `assert` is a bit misleading, since assertions usually used to throw errors. Maybe `as_shiny_tag`?
* This function will also change a string into a tagList containing the string. From prior discussion, I think we were trying to avoid that.
* I guess my broader question is, what is the purpose of this function?

R/tags.R
```diff
@@ -687,18 +733,40 @@ findDependencies <- function(tags, tagify = TRUE) {
...
+#'
+#' Resolves any [tagFunction()]s inside a list of [htmlDependencies()]. To
+#' resolve [tagFunction()]s _and then_ remove redundant dependencies all at once,
+#' use [resolveDependencies()] (which calls this function internally).
+#'
+#' @inheritParams resolveDependencies
+#' @export
+#' @return a list of HTML dependencies.
+#' @seealso [tagFunction()], [resolveDependencies()]
+resolveFunctionalDependencies <- function(dependencies) {
```
wch, line 758: `dependencies` will be `NULL` in the vast majority of cases, so it makes sense to have a check for that (an early return) at the top of this function.

R/html_dependency.R
```diff
@@ -160,28 +160,54 @@ htmlDependencies <- function(x) {
...
-
   } else {
     htmlDependencies(x) <- value
   }
+
   return(x)
 }
 
+
+asDependencies <- function(x) {
```
wch, line 183: It would help to have a comment saying that this will _not_ execute `tagFunction `s and convert them into static dependencies.

R/tags.R
```diff
@@ -687,18 +733,40 @@ findDependencies <- function(tags, tagify = TRUE) {
...
+  dependencies <- asDependencies(dependencies)
+  dependencies <- lapply(dependencies, function(dep) {
+    if (is_tag_function(dep)) {
+      dep <- dep()
+    }
+    if (isTag(dep) || isTagList(dep)) {
```
wch, line 764: Do we need this `if` code path? I thought that that we could assume that whatever is going into this function is a list of `htmlDependency`s and/or functions returning `htmlDependency`s -- and that no tag objects were allowed.

R/tags.R
```diff
@@ -907,10 +975,11 @@ as.tags <- function(x, ...) {
 
 #' @export
 as.tags.default <- function(x, ...) {
-  if (is.list(x) && !isTagList(x))
-    unclass(x)
-  else
+  if (is.list(x)) {
```
wch, line 978: There's an `as.tags.list` method now, so when would we hit this code path?

cpsievert: In this case: `as.tags(structure(list(), class = "foo"))`

I'll make sure to add a test that we handle this properly

R/tags.R
cpsievert:
```suggestion
  # Only non-classed lists will hit this method (classed lists will fall-through to the default method)
  do.call(tagList, x)
```

R/tags.R
wch:
```suggestion
  # Plain (non-classed) lists will hit as.tags.list(), but lists with a class
  # will get here. (tagLists will already have been handled by
  # as.tags.shiny.tag.list)
  if (is.list(x)) {
```

R/tags.R
cpsievert: I just remembered that (after our discussion), that this case is actually very important because of

```r
#' @export
as.tags.html_dependency <- function(x, ...) {
  attachDependencies(tagList(), x)
}
```

R/tags.R
cpsievert: Good to know!

R/tags.R
wch: A comment here explaining this would help future us (or whoever's working on this code).

Also, if our tests don't already cover this, we should add ones that do.

cpsievert: Actually, I don’t think we need the case at all since we should always be receiving an object that has passed through `htmlDependencies(x)`, so I added a warning in https://github.com/rstudio/htmltools/pull/180/commits/d9ca877a73cd175de3371c75f375199979818dff

wch (approved):

PR #183, opened 2020-09-29
Add unit tests for DOM tree order traversal (with tagFunction()s)

cpsievert:
Add tests related to assumptions made in https://github.com/rstudio/shiny/pull/3062

### DIFF
--- a/tests/testthat/test-tags.r
+++ b/tests/testthat/test-tags.r
@@ -530,12 +530,6 @@ test_that("Tag list tree is rendered in DOM tree order", {
   # Tree order is preorder, depth-first traversal
   # https://dom.spec.whatwg.org/#concept-tree
   #
-  # Test for depth-first (not breadth-first)
-  # https://stackoverflow.com/a/16113998/1583084
-  expect_identical(
-    as.character(tags$p("Consider this", tags$i("relatively"), "simple", tags$strong("HTML"), "subtree")),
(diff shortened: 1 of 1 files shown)


tests/testthat/test-tags.r
```diff
@@ -526,6 +526,37 @@ test_that("NA attributes are rendered correctly", {
   )
 })
 
+test_that("Tag list tree is rendered in DOM tree order", {
+  # Tree order is preorder, depth-first traversal
+  # https://dom.spec.whatwg.org/#concept-tree
+  #
+  # Test for depth-first (not breadth-first)
+  # https://stackoverflow.com/a/16113998/1583084
+  expect_identical(
```
wch, line 535: I don't see how this particular test checks how the tree is traversed. I think this test can be removed because the other test does check how the tree is traversed.

wch: Example of how tree traversal order shouldn't affect the output in a case like this:
https://gist.github.com/wch/5e4a5df737d0fbe1effca5621bea8dbb

cpsievert: Right -- I suppose this isn't testing `tagify()`/`rewriteTags()` tree traversal, but it is testing `flattenTags()`'s tree flattening algorithm.

Anyway, I'm happy to remove it since we have plenty other tests that will make sure that `flattenTags()` is depth-first

wch (approved):

PR #185, opened 2020-10-02
Allow to set the lang attribute, will close #82

### LINKED ISSUE #82: <html>after <!DOCTYPE html> tag should use lang="en" property/attribute
https://github.com/rstudio/htmltools/blob/master/R/html_print.R#L89
and please make it customizable e.g. lang="cs", lang="de", lang="lu" so that `shiny` can take advantage of it. 

BIG THANKS

ColinFay:


cpsievert:
Nice, thanks! We'd also appreciate a NEWS item and a unit test!

wch:
It seems to me that the `lang` should be a property of the object being saved, rather than an argument to `save_html()`. On the other hand, there's not an obvious place to store that property.

We currently have a PR for this in Shiny, where we're adding a `lang` attribute to the top-level tag in the `ui` object. However, that approach may not make as much sense for the general case that this PR is trying to address. https://github.com/rstudio/shiny/pull/2920

@yihui, @cderv, What cases do you have in mind where `save_html()` is called from?

ColinFay:
Test and NEWS bullet added ✅

tests/testthat/test-print.R
```diff
@@ -62,3 +62,20 @@ test_that("Special characters are not re-encoded", {
...
+  output_read <- readLines(output)
+  expect_true(
+    grepl("<html lang=\"en\">", paste(output_read, collapse = " "))
+  )
+  # test for fr
+  save_html("<h2>Howdy</h2>", output, lang = "fr")
+  output_read <- readLines(output)
+  expect_true(
+    grepl("<html lang=\"fr\">", paste(output_read, collapse = " "))
+  )
```
cpsievert, line 80: Let's also make sure `lang` can be overridden via `<body>` (in case the end user doesn't have control over `save_html()`) 

```suggestion
  )
  save_html('<body lang="fr">Howdy</body>', output)
  output_read <- readLines(output)
  expect_true(
    grepl("<body lang=\"fr\">", paste(output_read, collapse = " "))
  )
```

cderv:
> What cases do you have in mind where save_html() is called from?

I don't have a specific case in mind other than exporting some HTML to a file - it is just that I recently understand than setting `lang` attributes was always a good practice (and even a must-have) for any html file (especially for accessibility with screen reader). It seems good to me that `save_html` allows to create a html file with a correct `lang` attribute, as it takes care of creating the necessary `<html>` for the user. As `lang` attributes is part of  the whole document definition (`<html>` node), it seems ok to me to have it as a parameter. 

I am not sure anyone will provide `<html lang="fr">` in the content pass to `save_wiget`. And even if `htmltools::tags$html(lang="fr", ...)` is provided in the content,  it currently don't play well with the `save_html` function as  it would  generate an odd html structure, with a `<html>` tag in the middle of the document that the browser seems to recognize anyway... 

<details><summary> Example of providing `tags$html(lang = "fr", ...)` </summary>

``` r
library(htmltools)
temp_html <- tempfile(fileext = ".html")
html <-  tags$html(
  lang="fr", 
  tags$head(
    tags$style("h1{color:red;}")
  ),
  tags$body(
    tags$h1("title", id = "title1")
  )
)
save_html(html, temp_html)
xfun::file_string(temp_html)
#> <!DOCTYPE html>
#> <html>
#> <head>
#> <meta charset="utf-8"/>
#> <style>body{background-color:white;}</style>
#> 
#>   <style>h1{color:red;}</style>
#> </head>
#> <html lang="fr">
#>   <body>
#>     <h1 id="title1">title</h1>
#>   </body>
#> </html>
#> </html>
unlink(temp_html)
```

<sup>Created on 2020-10-05 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0.9001)</sup>

</details>

If we don't want to have specific attributes as function parameters, we could make it more generic and the function could take a list of `html.attr` to the `<html>` node created.
```r
save_html(html, file = "test.html", html.attr = list(lang="fr"))
```

tests/testthat/test-print.R
wch: Now that I think about it, I don't think we need to hoist a body tag's `lang` attribute -- if the user does something like `save_html('<body lang="fr">Howdy</body>', output)`, then my understanding is that the browser will interpret everything in the body to be in French. The same is true for any top-level tag: the content inside will be interpreted using the specified `lang`.

wch (approved):

PR #187, opened 2020-10-16
(SECURITY) Allow use of "integrity" and "crossorigin" attributes when specifying script dependencies

### LINKED ISSUE #101: Add support for <script> "integrity" and "crossorigin" attributes
In use of htmltools::htmlDependency() for CDNs, it would allow for SRI checking (which ensures resources hosted on third-party servers have not been tampered with), considered a best practice.

Example from jQuery:
```html
<script
  src="https://code.jquery.com/jquery-1.12.4.min.js"
  integrity="sha256-ZosEbRLbNQzLpnKIkEdrPv7lOy9C27hHQ+Xp8a4MxAQ="
  crossorigin="anonymous"></script>
```

References:
* [SRI](https://www.srihash.org/)
* [Browser support](https://caniuse.com/#feat=subresource-integrity)
* [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script)

matthewstrasiotto:
Resolves #101 - Use of such attributes are considered best-practise for security when referencing scripts stored at remote servers: 

https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity

### DIFF
--- a/R/html_dependency.R
+++ b/R/html_dependency.R
@@ -457,8 +457,7 @@ renderDependencies <- function(dependencies,
         " does not have a usable source")
 
     if (length(dep$script) > 1) {
-      usableScript <- scriptFields[which(scriptFields %in% names(dep$script))]
-      if (length(usableScript) == 0)
+      if (! all(names(script) %in% scriptFields))
         stop("Dependency ", dep$name, " ", dep$version,
           " does not have usable script fields")
     }
--- a/R/html_dependency.R
+++ b/R/html_dependency.R
@@ -36,6 +36,13 @@
 #'   \code{href} for URL. For example, a dependency that was both on disk and at
 #'   a URL might use \code{src = c(file=filepath, href=url)}.
 #'
+#'   \code{script} can be given as either a scalar string, or a named list with
+#'   the following fields: \code{src}, \code{integrity}, & \code{crossorigin},
+#'   allowing the use of SRI to ensure the integrity of packages downloaded from
+#'   remote servers.
+#'   \code{script = "pkg.js"} is equivalent to
+#'   \code{script = list(src = "pkg.js")}.
+#'
 #'   \code{attachment} can be used to make the indicated files available to the
 #'   JavaScript on the page via URL. For each element of \code{attachment}, an
 #'   element \code{<link id="DEPNAME-ATTACHINDEX-attachment" rel="attachment"
@@ -433,6 +440,7 @@ makeDependencyRelative <- function(dependency, basepath, mustWork = TRUE) {
 #' @export
 renderDependencies <- function(dependencies,
   srcType = c("href", "file"),
+  scriptFields = c("src", "integrity", "crossorigin"),
   encodeFunc = urlEncodePath,
   hrefFilter = identity) {
 
@@ -444,6 +452,13 @@ renderDependencies <- function(dependencies,
     if (length(usableType) == 0)
       stop("Dependency ", dep$name, " ", dep$version,
         " does not have a usable source")
+
+    if (length(dep$script) > 1) {
+      usableScript <- scriptFields[which(scriptFields %in% names(dep$script))]
+      if (length(usableScript) == 0)
+        stop("Dependency ", dep$name, " ", dep$version,
+          " does not have usable script fields")
+    }
 
     dir <- dep$src[head(usableType, 1)]
 
@@ -478,10 +493,37 @@ renderDependencies <- function(dependencies,
 
     # add scripts
     if (length(dep$script) > 0) {
+      if (length(dep$script) == 1) dep$script <- list(src = dep$script[[1]])
+
+      dep$script$src <- paste(
+        ' src="',
+        htmlEscape(hrefFilter(file.path(srcpath, encodeFunc(dep$script$src)))),
+        '"',
+        sep = ''
+      )
+
+      if (!is.null(dep$script$integrity)) {
+        dep$script$integrity <- paste(
+          ' integrity="', dep$script$integrity, '"',
+          sep = ''
+        )
+      }
+
+      if (!is.null(dep$script$crossorigin)) {
+        dep$script$crossorigin <- paste(
+          ' crossorigin="', htmlEscape(dep$script$crossorigin), '"',
+          sep = ''
+        )
+      } else {
+        dep$script$crossorigin <- ""
+      }
+
       html <- c(html, paste(
-        "<script src=\"",
-        htmlEscape(hrefFilter(file.path(srcpath, encodeFunc(dep$script)))),
-        "\"></script>",
+        "<script",
+        dep$script$src,
+        dep$script$integrity,
+        dep$script$crossorigin,
+        "></script>",
         sep = ""
       ))
     }
--- a/R/html_dependency.R
+++ b/R/html_dependency.R
@@ -521,27 +521,12 @@ renderDependencies <- function(dependencies,
 
 
 renderScript <- function(script, srcpath, encodeFunc, hrefFilter) {
-  leaf_fields <- c("src", "integrity", "crossorigin")
+  # If the input is a named list, transform it to an unnamed list
+  # whose only element is the input list
+  if (anyNamed(script)) script <- list(script)
 
-  check_names <- function(script) {
-    if (length(names(script)) != 0) {
-      if (!all(names(script) %in% leaf_fields)) {
-        invalid_names <- names(script)[!names(script) %in% leaf_fields]
-        stop(
-          "Providing ", paste(invalid_names, collapse = T), " not supported",
-          " for <script> tags")
-      }
-
-      if (!hasName(script, "src")) {
-        stop("Named script elements must provide src")
-      }
-    }
-  }
-
-  check_names(script)
-
-  if (length(names(script)) > 0) script <- list(script)
-
+  # For each element, if it's a scalar string, transform it to a named
+  # list with one element, "src".
   script <- lapply(script, function(item) {
     if (length(item) == 1 && is.character(item)) {
       item = list(src = item)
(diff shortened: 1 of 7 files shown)


R/html_dependency.R
```diff
@@ -433,6 +443,7 @@ makeDependencyRelative <- function(dependency, basepath, mustWork = TRUE) {
 #' @export
 renderDependencies <- function(dependencies,
   srcType = c("href", "file"),
+  scriptFields = c("src", "integrity", "crossorigin"),
```
wch, line 446: I don't think there's a need for this argument -- if these fields are provided, it might as well be used.

wch:
For `htmlDependency()`, the value of the `script` can currently be a vector of multiple JS files, as in:

```R
htmlDependency(
  name = "lib2",
  version = "1.0",
  src = list(href = "https://cdn.com/libs"), 
  script = c(
    "js/lib2.js",
    "js/lib2-extra.js"
  )
)
```

Here's a set of cases that would need to work:

```R
renderDependencies(
  list(
    # Single .js file with integrity and crossorigin
    htmlDependency(
      name = "lib1",
      version = "1.0",
      src = list(href = "https://cdn.com/libs"), 
      script = c(
        "js/lib1.js",
        integrity = "asdf",
        crossorigin = "anonymous"
      )
    ),

    # Multiple .js files
    htmlDependency(
      name = "lib2",
      version = "1.0",
      src = list(href = "https://cdn.com/libs"), 
      script = c(
        "js/lib2.js",
        "js/lib2-extra.js"
      )
    ),

    # Multiple .js files, one with integrity, crossorigin
    htmlDependency(
      name = "lib3",
      version = "1.0",
      src = list(href = "https://cdn.com/libs"), 
      script = list(
        c(
          "js/lib3.js",
          integrity = "asdf",
          crossorigin = "anonymous"
        ),
        "js/lib3-extra.js"
      )
    ),
    
    # Multiple .js files, all with integrity, crossorigin;
    # one in list, one in character vector
    htmlDependency(
      name = "lib4",
      version = "1.0",
      src = list(href = "https://cdn.com/libs"), 
      script = list(
        list(
          "js/lib4.js",
          integrity = "asdf",
          crossorigin = "anonymous"
        ),
        c(
          "js/lib4-extra.js",
          integrity = "asdf",
          crossorigin = "anonymous"
        )
      )
    )
  )
)
```

It would also be helpful to have examples in the docs that illustrate how to provide `integrity` and `crossorigin`. (The docs for `htmlDependency` don't currently have any examples, but it would benefit from having some.)

R/html_dependency.R
```diff
@@ -478,10 +496,37 @@ renderDependencies <- function(dependencies,
...
+        dep$script$crossorigin <- ""
+      }
+
       html <- c(html, paste(
-        "<script src=\"",
-        htmlEscape(hrefFilter(file.path(srcpath, encodeFunc(dep$script)))),
-        "\"></script>",
+        "<script",
+        dep$script$src,
+        dep$script$integrity,
```
wch, line 527: Instead of the `if` statements up above, this could simply be:

```R
if (!is.null(dep$script$integrity)) paste0(' integrity="', dep$script$integrity, '"')
```

And similar for `crossorigin`.

Also, modifying the `dep` object in place and replacing the original values with these fragments of HTML doesn't feel right.

R/html_dependency.R
```diff
@@ -478,10 +496,37 @@ renderDependencies <- function(dependencies,
 
     # add scripts
     if (length(dep$script) > 0) {
+      if (length(dep$script) == 1) dep$script <- list(src = dep$script[[1]])
```
wch, line 498: It will probably be best to extract this script handling code into a function (which can be defined within `renderDependencies`, since the input value for `dep$script` could either represent a single object with a structure like `list(src="a.js", integrity="abc")`, or a list of such objects. (And additionally, it must handle values like `list("a.js", "b.js")`.

R/html_dependency.R
matthewstrasiotto: > Also, modifying the dep object in place and replacing the original values with these fragments of HTML doesn't feel right.

Yeah I felt that too, I think I might refactor this out into a function anyway, though, because `<script>` isn't the only tag that can use these attributes, it's just the most important for `<script>`

matthewstrasiotto:
Thanks for the quick response, & for the feedback - I already discovered that the `script` can be a character vector the hard way, when I was trying to use my fork w/ `kableExtra` & got errors on its bootstrap dep.

matthewstrasiotto:
It sounds like the way forward is as follows: 

For each element  `item` of `script`:
  if `item` is a list 
    check `item` for names & transform into attributes
  else 
    put `item` directly into script tags

matthewstrasiotto:
For my documentation of the script tag- I don't have time to look up the Rd spec on writing lists, and I usually use markdown for my docs, so the lists will need to be reformated, but this is ready to pull besides that @wch

R/html_dependency.R
```diff
@@ -505,6 +518,68 @@ renderDependencies <- function(dependencies,
   HTML(paste(html, collapse = "\n"))
 }
 
+
+
+renderScript <- function(script, srcpath, encodeFunc, hrefFilter) {
+  leaf_fields <- c("src", "integrity", "crossorigin")
+
+  check_names <- function(script) {
```
wch, line 526: Is there a reason to limit the attributes to `src`, `integrity`, and `crossorigin`? I think we can just remove the name check.

R/html_dependency.R
```diff
@@ -505,6 +518,68 @@ renderDependencies <- function(dependencies,
...
+
+      if (!hasName(script, "src")) {
+        stop("Named script elements must provide src")
+      }
+    }
+  }
+
+  check_names(script)
+
+  if (length(names(script)) > 0) script <- list(script)
```
wch, line 543: I think the algorithm can go something like this:

* If the input is all unnamed, like `c("a.js", "b.js")`, convert to `list(list(src="a.js"), list(src="b.js"))` with something like:
    ```R
    lapply(script, function(x) list(src = x))
    ```
* If all named, skip that step.
* If it’s a mix of named and unnamed, throw an error. 


Keep in mind that vectors/lists can have names, but the names are empty. For example:

```R
# Named list with empty names:
x <- list(a=1, 2, 3)[2:3]
names(x)
#> [1] "" ""

# Compare to truly unnamed list
names(list(2, 3))
#> NULL
```

Note that in Shiny, we have functions that take this into account. It might make sense to bring those (or similar functions) into htmltools.
https://github.com/rstudio/shiny/blob/7aa3a243ba7ee33c89060a52cb7518e892421ecf/R/utils.R#L147-L174

wch (changes requested):
In addition to the other comments, I think there also be some tests for non-list input, to ensure that it's handled correctly.

R/html_dependency.R
```diff
@@ -505,6 +518,68 @@ renderDependencies <- function(dependencies,
   HTML(paste(html, collapse = "\n"))
 }
 
+
+
+renderScript <- function(script, srcpath, encodeFunc, hrefFilter) {
```
cpsievert, line 523: To make this change safer and easier to reason about, consider adding an early exit here, something like:

```r
if (is.character(script)) {
  return(renderScriptSrc(script, srcpath, encodeFunc, hrefFilter))
}
```

With `renderScriptSrc()` defined as:

```r
renderScriptSrc <- function(script, path, encodeFunc, hrefFilter) {
  paste0(
    '<script src="',
    htmlEscape(hrefFilter(file.path(path, encodeFunc(script)))),
    "'></script>"
  )
}
```

You could then also reuse this function when you go to render the more complex list case

tests/testthat/test-deps.r
```diff
@@ -186,3 +186,121 @@ test_that("Modifying children using dependencies", {
...
+      file.path(src, nm),
+      file.path(src, nm2)
+      )
+
+    expect <- paste(
+      '<script src="', src_urls[[1]],'"></script>\n',
+      '<script src="', src_urls[[2]],'"></script>',
+      sep = "")
+
+    expect_equal(!!as.character(out), !!expect)
```
wch, line 289: Why are the `!!` needed?

matthewstrasiotto: !! Lead to eager evaluation in test that output reports, making it easier to understand what the actual diffs are

Not sure when it was added, so if it isn't backwards compatible, sorry about that!

R/html_dependency.R
```diff
@@ -505,6 +518,68 @@ renderDependencies <- function(dependencies,
...
+    }
+    check_names(item)
+
+    return(item)
+  })
+
+  script <- vapply(
+    script,
+    function(x) {
+      z <- is.null
```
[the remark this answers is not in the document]
matthewstrasiotto: Yeah, i should have known that would bug people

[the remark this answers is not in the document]
matthewstrasiotto: I'll probably refactor this into it's own function & iterate directly over the names of the script, given https://github.com/rstudio/htmltools/pull/187#discussion_r507870443

R/html_dependency.R
matthewstrasiotto: I don't really see how that could be used for the more complex list case.

The llist case aims to support instances such as:

```r
renderScript(list("s1.js", list(src = "s2.js"), list(src = "s3.js", integrity = "hash", crossorigin = "anonymous")))
```

```html
<script src="s1.js"></script>
<script src="s2.js"></script>
<script src="s3.js" integrity="hash" crossorigin="anonymous"></script>
```

or:

```r
renderScript(list(src = "s3.js", integrity = "hash", crossorigin = "anonymous"))
```

```html
<script src="s3.js" integrity="hash" crossorigin="anonymous"></script>
```

These accept mixed use of character vectors, nested lists, & pure lists.

Unless I'm mistaken, the change you're proposing would mean I'd have to duplicate the actual templating pattern for the case of unnamed character vectors, and named lists.

R/html_dependency.R
matthewstrasiotto: Would you like me to do so in this PR?

matthewstrasiotto: I've added in the two mentioned utils to `utils.R`.

It accepts a named list of characters, or an unamed list of lists/characters. I forgot to throw if it's a mix (oops)

matthewstrasiotto:
@wch @cpsievert the requested changes have been addressed, except for https://github.com/rstudio/htmltools/pull/187#discussion_r507898122 , on which I disagree (or perhaps misunderstood) the requested change. 

Should be ready for pull, let me know if you'd like me to squash some of the history down.

matthewstrasiotto:
Cleaned & reopened this in #188

PR #188, opened 2020-10-23
(SECURITY) Allow use of "integrity", "crossorigin" as well as any other attributes in htmlDependency script parameter

### LINKED ISSUE #101: Add support for <script> "integrity" and "crossorigin" attributes

matthewstrasiotto:
This is a cleaned up duplicate of #187 :

Resolves #101 - Use of such attributes are considered best-practise for security when referencing scripts stored at remote servers: 

https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity

Resolves #101 - Use of such attributes are considered best-practise for security when referencing scripts stored at remote servers: 

https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity

From: https://github.com/rstudio/htmltools/pull/187#issuecomment-710118335

> For `htmlDependency()`, the value of the `script` can currently be a vector of multiple JS files, as in:
> 
> ```r
> htmlDependency(
>   name = "lib2",
>   version = "1.0",
>   src = list(href = "https://cdn.com/libs"), 
>   script = c(
>     "js/lib2.js",
>     "js/lib2-extra.js"
>   )
> )
> ```

> Here's a set of cases that would need to work:

> ```r
> renderDependencies(
>   list(
>     # Single .js file with integrity and crossorigin
>     htmlDependency(
>       name = "lib1",
>       version = "1.0",
>       src = list(href = "https://cdn.com/libs"), 
>       script = c(
>         "js/lib1.js",
>         integrity = "asdf",
>         crossorigin = "anonymous"
>       )
>     ),
>     # Multiple .js files
>     htmlDependency(
>       name = "lib2",
>       version = "1.0",
>       src = list(href = "https://cdn.com/libs"), 
>       script = c(
>         "js/lib2.js",
>         "js/lib2-extra.js"
>       )
>     ),
> 
>     # Multiple .js files, one with integrity, crossorigin
>     htmlDependency(...) # (omitted from quote)
>     
>     # Multiple .js files, all with integrity, crossorigin;
>     # one in list, one in character vector
>     htmlDependency(...) # (omitted from quote)
> ```

I have implemented the requested cases, though with some key differences: 

I have chosen **NOT** to support the case of providing a complex script tag in the following format:

```r
htmlDependency(
  src = "cdnjs.com/libraries/mypkg",
  script = list(
    "implicit_src_attribute.js", # Mixing and matching unnamed & named attributes is NOT supported
    integrity = "explicit named attribute"
  )
)
```

```r
htmlDependency(
  src = "cdnjs.com/libraries/mypkg",
  script = list(
     list(
      "implicit_src_attribute.js", # Mixing and matching unnamed & named attributes is NOT supported,
      integrity = "etcetera"
    ),
      "implicit_src_attribute2.js" # If it's a scalar, unnamed string, implic src IS supported
  )
)
```

If users are making the effort to dig up the SRI hashes for their scripts from their CDNs, and specifying them explicitly, I think it's valid to expect that they're also explicity about which field is their `src`. 

> It would also be helpful to have examples in the docs that illustrate how to provide `integrity` and `crossorigin`. (The docs for `htmlDependency` don't currently have any examples, but it would benefit from having some.)

I've documented this at the following:

https://github.com/rstudio/htmltools/blob/f2f86efb8495b93547994a374914231bbe269adf/R/html_dependency.R#L39-L62

### DIFF
--- a/R/html_dependency.R
+++ b/R/html_dependency.R
@@ -36,6 +36,31 @@
 #'   \code{href} for URL. For example, a dependency that was both on disk and at
 #'   a URL might use \code{src = c(file=filepath, href=url)}.
 #'
+#'   \code{script} can be given as one of the following:
+#'   \itemize{
+#'   \item a character vector specifying various scripts to include relative to the
+#'     value of \code{src}.
+#'     Each is expanded into its own \code{<script>} tag
+#'   \item A named list with any of the following fields:
+#'   \itemize{
+#'     \item \code{src},
+#'     \item \code{integrity}, &
+#'     \item \code{crossorigin},
+#'     \item any other valid \code{<script>} attributes.
+#'     }
+#'     allowing the use of SRI to ensure the integrity of packages downloaded from
+#'     remote servers.
+#'     Eg: \code{script = list(src = "min.js", integrity = "hash")}
+#'   \item An unamed list, containing a combination of named list with the fields
+#'     mentioned previously, and strings.
+#'     Eg:
+#'     \itemize{
+#'     \item \code{script = list(list(src = "min.js"), "util.js", list(src = "log.js"))}
+#'     \item \code{script = "pkg.js"} is equivalent to
+#'     \item \code{script = list(src = "pkg.js")}.
+#'     }
+#'   }
+#'
 #'   \code{attachment} can be used to make the indicated files available to the
 #'   JavaScript on the page via URL. For each element of \code{attachment}, an
 #'   element \code{<link id="DEPNAME-ATTACHINDEX-attachment" rel="attachment"
@@ -478,12 +503,7 @@ renderDependencies <- function(dependencies,
 
     # add scripts
     if (length(dep$script) > 0) {
-      html <- c(html, paste(
-        "<script src=\"",
-        htmlEscape(hrefFilter(file.path(srcpath, encodeFunc(dep$script)))),
-        "\"></script>",
-        sep = ""
-      ))
+      html <- c(html, renderScript(dep$script, srcpath, encodeFunc, hrefFilter))
     }
 
     if (length(dep$attachment) > 0) {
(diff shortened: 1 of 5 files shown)


wch (changes requested):
I agree that if they are providing any named attributes, `src` should also be named.

Have you signed a contributor agreement?
https://github.com/rstudio/htmltools/blob/master/CONTRIBUTING.md

matthewstrasiotto:
> I agree that if they are providing any named attributes, `src` should also be named.
> 
> Have you signed a contributor agreement?
> https://github.com/rstudio/htmltools/blob/master/CONTRIBUTING.md

Signed and sent now 🙂

R/html_dependency.R
```diff
@@ -505,6 +525,96 @@ renderDependencies <- function(dependencies,
...
+    return(item)
+  })
+
+  script <- vapply(
+    script, function(x) {
+      x$src <- hrefFilter(file.path(srcpath, encodeFunc(x$src)))
+      renderTag(x, tag_name = "script", required_attributes = "src")
+    },
+    FUN.VALUE = character(1)
+    )
```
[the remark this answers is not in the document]
matthewstrasiotto: So do i! I missed that

R/html_dependency.R
```diff
@@ -505,6 +525,96 @@ renderDependencies <- function(dependencies,
...
+      )
+    }
+
+    return(item)
+  })
+
+  script <- vapply(
+    script, function(x) {
+      x$src <- hrefFilter(file.path(srcpath, encodeFunc(x$src)))
+      renderTag(x, tag_name = "script", required_attributes = "src")
```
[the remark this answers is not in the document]
matthewstrasiotto: I'm not sure i agree, i think this could be useful if the other tag renders were modified to call this for more DRY, but I'm happy to remove it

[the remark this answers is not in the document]
wch: Yeah, I think maintaining a separate `renderTag` function isn't worth it here, especially since it's fairly straightforward to understand what the `paste` calls are doing here.

matthewstrasiotto:
Requested changes made, rtp

R/html_dependency.R
```diff
@@ -505,6 +525,50 @@ renderDependencies <- function(dependencies,
...
+
+    return(item)
+  })
+
+  script <- vapply(
+    script, function(x) {
+      x$src <- hrefFilter(file.path(srcpath, encodeFunc(x$src)))
+      paste0(
+        "<script", 
+        paste0(" ", names(x), '="', x, '"', collapse = ''),
```
wch, line 560: Sorry, I forgot to mention that there should be `htmlEscape(x)` here.

R/html_dependency.R
```diff
@@ -505,6 +525,50 @@ renderDependencies <- function(dependencies,
...
+       "></script>",
+       collapse = ""
+       )
```
wch, line 563: Indentation nit again :)

R/html_dependency.R
wch: It would be good if one of the tests had a character that needs escaping, to make sure this doesn't get overlooked if there are changes in the future.

R/html_dependency.R
matthewstrasiotto: youre kidding! Im editing on my phone via ssh and vim so i might be best off waiting til I'm in front of a keyboard

matthewstrasiotto: Should be resolved for real this time

R/html_dependency.R
matthewstrasiotto: I think those test cases are a very good idea, it's easy behavior to regress and dangerous if allowed to happen, 

however i think they constitute a good chunk of testing, and kind of comprise their own unit of testing, so i wont be adding that to this PR

matthewstrasiotto: Changes implemented

wch: FWIW, I think just that if you modify one test case with a weird character, that's probably sufficient for our purposes.

matthewstrasiotto: Resolved, added a brand new test case with an explicit purpose though

matthewstrasiotto:
Requested changes have been made @wch

wch (approved):

wch:
Looks good, thanks!

PR #192, opened 2020-11-19
save_html() now works with a subdirectory file or connection

cpsievert:
Supercedes #104. This also makes `save_html()` work with  `file()` connections, which was for some reason partially addressed in https://github.com/rstudio/htmltools/commit/3dba0188715f09883f9efb1bf46956a6e1f823f8, but I don't think that has ever worked (until now)

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -7,6 +7,8 @@ htmltools 0.5.0.9002
 
 * `save_html()` now has a `lang` parameter that can be used to set the lang attribute of `<html>`. (@ColinFay, #185)
 
+* `save_html()`'s `file` argument now properly handles relative paths. (@haozhu233, #105)
+
 * Closed #101: `htmlDependency` & `renderDependencies` now allow the `script` argument to be given as a named list containing the elements: `src`, `integrity`, `crossorigin`. (@matthewstrasiotto, #188)
 
 * Closed #189: `validateCssUnit()` now accepts `fit-content`. (#190)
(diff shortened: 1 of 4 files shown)


NEWS
```diff
@@ -7,6 +7,8 @@ htmltools 0.5.0.9002
 
 * `save_html()` now has a `lang` parameter that can be used to set the lang attribute of `<html>`. (@ColinFay, #185)
 
+* Closed #104: `save_html()`'s `file` argument now properly handles relative paths. (@haozhu233, #105)
```
wch, line 10:
```suggestion
* Closed #104: `save_html()`'s `file` argument now properly handles relative paths. (@haozhu233, #105, #192)
```

schloerke (approved):

R/html_print.R
```diff
@@ -76,9 +76,12 @@ save_html <- function(html, file, background = "white", libdir = "lib", lang = "
 
   # ensure that the paths to dependencies are relative to the base
   # directory where the webpage is being built.
-  dir <- dirname(file)
-  oldwd <- setwd(dir)
-  on.exit(setwd(oldwd), add = TRUE)
+  if (is.character(file)) {
+    dir <- normalizePath(dirname(file), mustWork = TRUE)
```
wch, line 80: I'd also suggest changing the docs:

```
#' @param file File to write content to. The containing directory must already exist must already exist.
```

PR #194, opened 2020-12-02
parseCssColors(): treat incoming NAs as an invalid color

### LINKED ISSUE #161: parseCssColors errors on missing value
It should probably return NA instead

```r
htmltools::parseCssColors(NA)
```

```r
Error in substring(u, so, so + ml - 1L) : invalid substring arguments
```

```r
htmltools::parseCssColors(c("red", NA))
```

```r
Error in substring(u, so, so + ml - 1L) : invalid substring arguments
```

cpsievert:
Closes #161

### DIFF
--- a/R/colors.R
+++ b/R/colors.R
@@ -34,10 +34,11 @@ parseCssColors <- function(str, mustWork = TRUE) {
   # Note that an empty string is not a valid color, so parsing fails
   # on NA input values, and thus, will be converted back to NA
   # when `mustWork = FALSE`
-  if (!(is.character(str) || rlang::is_na(str))) {
+  isNA <- is.na(str)
+  if (!(is.character(str) || all(isNA))) {
     stop("`str` must be a character vector (or NA).")
   }
-  str[is.na(str)] <- ""
+  str[isNA] <- ""
 
   # Strip insignificant whitespace
   str <- color_strip_ws(str)
--- a/R/colors.R
+++ b/R/colors.R
@@ -30,6 +30,13 @@
 #' @md
 #' @export
 parseCssColors <- function(str, mustWork = TRUE) {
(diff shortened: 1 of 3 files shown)


wch (approved):
Looks good! 
Can you also add a NEWS item?

tests/testthat/test-colors.R
```diff
@@ -119,3 +119,19 @@ test_that("decode_color_keyword", {
   expect_error(decode_color_keyword(" orange "))
   expect_error(decode_color_keyword(NA))
 })
+
+test_that("parseCssColors() handles incoming NA values sensibly", {
+  expect_error(parseCssColors(NA))
+  expect_identical(
+    parseCssColors(NA_character_, mustWork = FALSE),
```
wch, line 126: I think we should also keep the old version of the test, with `NA_logical_`, because it seems likely that users will pass in regular old `NA`. The logic for `parseCssColors` will also have to be updated to handle this.

R/colors.R
```diff
@@ -30,6 +30,15 @@
 #' @md
 #' @export
 parseCssColors <- function(str, mustWork = TRUE) {
+  # Logic below assumes a character string with non-missing values
+  # Note that an empty string is not a valid color, so parsing fails
+  # on NA input values, and thus, will be converted back to NA
+  # when `mustWork = FALSE`
+  if (!(is.character(str) || rlang::is_na(str))) {
```
wch, line 37: But what about the cases where it's a vector of (non-character) `NA`s? As in:

```R
parseCssColors(c(NA, NA))
```

cpsievert: For some reason I was thinking that'd be a character vector, but I suppose logical makes way more sense!

wch (approved):

PR #198, opened 2021-01-15
Render any NA values in <script> attributes as a boolean attribute

### LINKED ISSUE #197: Boolean attributes in script tag not rendered correctly
In #188, we added the ability to specify `<script>` attributes with a list of values. However, it doesn't seem to support boolean attribites in the same way as other tags (i.e., `div(foo=NA)` gets rendered as `<div foo></div>`). 

For example:

```r
library(htmltools)
dep <- htmlDependency(
  "foo", "1.0",
  src = list(href = "foo/"),
  script = list(src = "my.js", defer = NA)
)
renderDependencies(list(dep))
#> <script src="foo/my.js" defer="NA"></script>
```

BTW, the above seems to work in Chrome/Firefox, but it really should be rendered as 

```r
#> <script src="foo/my.js" defer></script>
```

cpsievert:
Closes #197

TODO: unit tests

### DIFF
--- a/R/html_dependency.R
+++ b/R/html_dependency.R
@@ -555,19 +555,10 @@ renderScript <- function(script, srcpath, encodeFunc, hrefFilter) {
   script <- vapply(
     script, function(x) {
       x$src <- hrefFilter(file.path(srcpath, encodeFunc(x$src)))
-      paste0(
-        "<script", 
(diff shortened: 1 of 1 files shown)


cpsievert:
Superceded by #201

PR #204, opened 2021-03-17
doRenderTags() now handles a top-level tagFunction()

### LINKED ISSUE #203: doRenderTags() fails with a top-level tagFunction()
```r
> library(htmltools)
> doRenderTags(tagFunction(function() { div() }))
Error: object of type 'closure' is not subsettable
```

cpsievert:
Closes #203

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,7 @@ htmltools 0.5.1.9000
 --------------------------------------------------------------------------------
 
 * Closed #197: Fixed rendering of boolean attributes in <script> tags rendered via renderDependencies() (#197, thanks @atusy).
+* Closed #204: doRenderTags() now works when provided a top-level tagFunction(). (#204)
 
 htmltools 0.5.1.1
 --------------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


NEWS.md
```diff
@@ -2,6 +2,7 @@ htmltools 0.5.1.9000
 --------------------------------------------------------------------------------
 
 * Closed #197: Fixed rendering of boolean attributes in <script> tags rendered via renderDependencies() (#197, thanks @atusy).
+* Closed #204: doRenderTags() now works when provided a top-level tagFunction(). (#204)
```
wch, line 5: Should be `Closed #203`

cpsievert: Thanks, fixed by c200de2d083cfda49949e0ae75abccd5f8d816ff

wch:
I'm a bit concerned that if tag functions are passed directly to `doRenderTags()`, information like html dependencies will be lost. In my reading of the code is that, until this change, `doRenderTags()` has been meant to be used with plain tag-like objects which have already had that processing done (usually by `renderTags()`). After this change, it's unclear to me exactly what `doRenderTags()` is supposed to accept -- it can now accept `tagFunction`s, but in so doing it will lose html dependency info.

When does this change actually makes a difference in behavior, and does it make more sense to fix those cases?

cpsievert:
Good point.

The motivation for this change is to make `shiny:::processDeps(buildTabItem(...))` (i.e., `insertTab()` et. al) work in https://github.com/rstudio/shiny/pull/3315.

I'll revert this change and add the special case in `processDeps()` instead

Issue #205, opened 2021-03-22
Tag classes should be normalized on creation


wch:
Tag classes are way too freeform and this makes it difficult to work with them. For example:

```R
x <- div(class = "a", class = "b c", class = c("d", "e f"))
str(x)
#> List of 3
#>  $ name    : chr "div"
#>  $ attribs :List of 3
#>   ..$ class: chr "a"
#>   ..$ class: chr "b c"
#>   ..$ class: chr [1:2] "d" "e f"
#>  $ children: list()
#>  - attr(*, "class")= chr "shiny.tag"
```

It would be better if this were translated to a data structure like `c("a", "b", "c", "d", "e", "f")` or `"a b c d e f"` when the tag is created.

Edit: I've realized that `class` isn't special; it makes sense to do this with all HTML attributes. We'll have to be careful about types, though (attributes can be atomic vectors, but they can also be lists or tag objects).

We should also make `tagAppendAttribute` and friends do the same thing.

schloerke:
This can not happen without major breaking changes to downstream dependencies. 

Instead, `tagGetAttribute(x, attr)` will be updated to return a list structure of any `attribs` value is not an atomic value.

PR #208, opened 2021-04-01
Add `tagQuery(tags)` for complex css selections and alterations

schloerke:
This implementation is experimental and will probably change.

---------------

`tagQuery(tags)` methods

### Select
* `$find(cssSelector)`: Find all tag elements matching the multi-element `cssSelector`
* `$children(cssSelector = NULL)`: Update the selected elements to the selected elements direct children.
* `$parent(cssSelector = NULL)`: Update the selected elements to the selected elements direct parents.
* `$parents(cssSelector = NULL)`: Update the selected elements to all ancestors of the selected elements.
* `$closest(cssSelector = NULL)`: Update the selected elements to closest ancestor (and including itself) of each of the selected elements.
* `$siblings(cssSelector = NULL)`: Get the siblings of each element in the set of matched elements.
* `$filter(fn)`: Update the selected elements if `fn(selected_i)` returns `TRUE`
* `$reset()`: Resets the selected elements to the top level (root) tags.

### Update

* `$addClass(class)`: Apps class(es) to each of the the selected elements.
* `$removeClass(class)`: Removes a set of class values from all of the selected elements.
* `$hasClass(class)`: Determine whether the selected elements have a given class. Returns a vector of logical values.
* `$toggleClass(class)`: If the a class value is missing, add it. If a  class value already exists, remove it.
* `$addAttrs(...)`: Add named attributes to all selected children. Similar to [`tagAppendAttributes()`].
* `$removeAttrs(attrs)`: Removes the provided attributes in each of the selected elements.
* `$emptyAttrs()`: Removes all attributes in each of the selected elements.
* `$hasAttr(attr)`: Returns a vector whose values are whether the selected element contains the non-`NULL` attribute.

### Adjust children
* `$append(...)`: Add all `...` objects as children **after** any existing children to the selected elements. Similar to [`tagAppendChildren()`]
* `$prepend(...)`: Add all `...` objects as children **before** any existing children to the selected elements. A variation of [`tagAppendChildren()`]
* `$empty()`: Remove all children in the selected elements. Use this method before calling `$append(...)` to replace all selected elements' children.

### Adjust siblings
* `$after(...)`: Add all `...` objects as siblings after each of the selected elements.
* `$before(...)`: Add all `...` objects as siblings before each of the selected elements.
* `$replaceWith(...)`: Replace all selected elements with `...`. This also sets the selected elements to an empty set.
* `$remove(...)`: Remove all selected elements from the `tagQuery()` object. The selected elements is set to an empty set.

## Generic methods
* `$each(fn)`: Perform function `fn` on each of the selected elements.

## Tag Query methods
* `$asTags(selected = TRUE)`: Converts all tag environments to tags.
* `$root()`: Return all top level (root) tags environments. 
* `$selected()`: Returns a list of selected tag environments.
* `$get(position)`: Returns the selected tag element at the position `position`.
* `$rebuild()`: Makes sure that all tags have been upgraded to tag environments. 
* `$print()`: Internal print method. Called by `print.htmltools.tag.query()`

----------

# Current tag env design decisions
* Tag graphs or tag environments can NOT be used in UI. These objects MUST be converted back to standard tag objects.
* tagFunctions will not be altered in place


R/selector.R
```diff
@@ -0,0 +1,233 @@
...
+  reg_info <- gregexpr(pattern, x, ...)
+  first <- reg_info[[1]]
+  if (length(first) == 1 && first == -1) {
+    return(NULL)
+  }
+
+  regmatches(x, reg_info)[[1]]
+}
+
+str_replace <- function(x, pattern, value, ...) {
```
wch, line 203: Can this be implemented with `sub()`?

R/selector.R
```diff
@@ -0,0 +1,233 @@
...
+  reg_info <- regexpr(pattern, x, ...)
+  if (length(reg_info) == 1 && reg_info == -1) {
+    return(x)
+  }
+
+  regmatches(x, reg_info) <- value
+  x
+}
+
+str_replace_all <- function(x, pattern, value, ...) {
```
wch, line 213: Implement with `gsub`

R/selector.R
wch: See https://github.com/wch/staticimports/blob/main/inst/staticexports/stringr.R

R/tag_env.R
```diff
@@ -0,0 +1,1307 @@
...
+  stopifnot(is.list(els))
+  walk_i(els, function(el, i) {
+    if (!isTagEnv(el)) {
+      stop("Object in position `", i, "` is not a tag environment")
+    }
+  })
+}
+
+# Return function that will verify elements before performing `func(els, fn)`
+selected_walk_gen <- function(func) {
```
wch, line 753: Should add `force(func)` at the top of this function.

R/tag_env.R
```diff
@@ -0,0 +1,1307 @@
...
+  )
+}
+# use for consistent `as_list()` order
+envir_stack <- function() {
+  stack <- faststack()
+  list(
+    push = stack$push,
+    as_list = stack$as_list,
+    unique_list = function() {
+      unique(stack$as_list())
```
wch, line 124: The `unique` is potentially expensive, although for most cases that we're dealing with, I think it's not an issue.

For cases where you want to do this, it might make sense to make a data structure that combines a `fastmap` (for fast duplicate checking when items are added) with a `faststack` (for preserving order). But we'll want to do benchmarks before we decide if that's needed. It could be that the overhead of creating both data structures is more costly than the `unique()` call.

R/tag_env.R
```diff
@@ -0,0 +1,1307 @@
...
+  x$env_key %||% stop("`x` is not a tag environment. `x$env_key` is missing")
+}
+
+# use for `has()` functionality
+envir_map <- function() {
+  map <- fastmap()
+  list(
+    keys = function() {
+      map$keys()
+    },
```
wch, line 102:
```suggestion
    keys = map$keys(),
```

schloerke: Maybe??

```suggestion
    keys = map$keys,
```

R/tag_env.R
```diff
@@ -0,0 +1,1308 @@
...
+}
+
+# Convert a list to an environment and keep class and attribute information
+safe_list_2_env <- function(x, new_class = NULL) {
+  x_list <- x
+
+  ret <- list2env(x_list, new.env(parent = emptyenv()))
+
+  attr_vals <- safe_attr_values(x_list)
+  Map(names(attr_vals), attr_vals, f = function(attr_name, attr_value) {
```
wch, line 152: Can you use `walk2` from https://github.com/wch/staticimports/blob/main/inst/staticexports/map.R ?

It will be slightly faster.

R/selector.R
```diff
@@ -0,0 +1,235 @@
...
+  if (length(reg_info) == 1 && reg_info == -1) {
+    return(NULL)
+  }
+
+  regmatches(x, reg_info)
+}
+
+# return a vector of matches or NULL
+str_match_all <- function(x, pattern, ...) {
+  stopifnot(length(x) == 1)
```
wch, line 221: Here and elsewhere, avoid `stopifnot` because it's slow.

```R
f1 <- function(x) {
  stopifnot(x > 0)
  NULL
}
f2 <- function(x) {
  if (!(x > 0)) stop()
  NULL
}
bench::mark(
  f1(1),
  f2(1)
)
#> # A tibble: 2 x 6
#>   expression      min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 f1(1)        1.92µs    2.7µs   253774.        0B        0
#> 2 f2(1)         271ns    323ns  1475025.        0B        0
```

R/tag_env.R
```diff
@@ -0,0 +1,1308 @@
...
+  })
+
+  class(ret) <- c(new_class, class(x_list), "environment")
+  ret
+}
+
+# Convert an environment to a list and keep class and attribute information
+safe_env_2_list <- function(x, old_class = NULL) {
+  x_env <- x
+  ret <- as.list.environment(x_env)
```
wch, line 163: Do we need `all.names = TRUE` here?

R/tag_env.R
wch: Oh, right.

R/tag_env.R
schloerke: Not now, but we will. I'll add it. Thanks!

R/tag_env.R
```diff
@@ -0,0 +1,1419 @@
...
+        #'
+        #'
+        #' * `$find(cssSelector)`: Find all tag elements matching the multi-element `cssSelector` starting from each selected element's children. If nothing has been selected, it will start from the root elements. The tag query object's selected elements will be updated with the matching set of tag environments.
+        find = function(cssSelector) {
+          rebuild()
+          setSelected(
+            tagQueryFindAll(getSelected(), cssSelector)
+          )
+          self
+        },
```
cpsievert, line 452: `x.find("div")` doesn't mutate `x` in jQuery...

R/tag_env.R
```diff
@@ -0,0 +1,1419 @@
...
+          rebuild()
+          self
+        },
+        #' * `$empty()`: Remove all children in the selected elements. Use this method before calling `$append(...)` to replace all selected elements' children.
+        empty = function() {
+          rebuild()
+          tagQueryChildrenEmpty(manuallySelected())
+          # no need to rebuild
+          self
+        },
```
cpsievert, line 581: Maybe these sort of functions (that operate on a selection) should warn when there is no selection (and inside devmode?)

R/tags.R
```diff
@@ -348,6 +348,12 @@ tagSetChildren <- function(tag, ..., list = NULL) {
   tag
 }
 
+tagInsertChildren <- function(tag, after, ..., list = NULL) {
+  throw_if_tag_function(tag)
+  tag$children <- unname(append(tag$children, c(dots_list(...), list), after))
+  tag
+}
```
cpsievert, line 355: Maybe worth exporting?

R/tags.R
```diff
@@ -934,8 +929,10 @@ tagify <- function(x) {
 
 # Given a list of tags, lists, and other items, return a flat list, where the
 # items from the inner, nested lists are pulled to the top level, recursively.
+# Be sure to check for tagEnvLike objects and not allow them
 flattenTags <- function(x) {
-  if (isTag(x)) {
+  assertNotTagEnvLike(x, "flattenTags")
+  if (isTag(x) || isTagEnv(x)) {
```
cpsievert, line 935:
```suggestion
  if (isTag(x)) {
```

R/tags.R
```diff
@@ -958,6 +954,47 @@ flattenTags <- function(x) {
     flattenTags(as.tags(x))
   }
 }
+flattenTagsRaw <- function(x) {
```
cpsievert, line 957: Add a comment explaining why we need a different version of `flattenTags`

R/tag_env.R
```diff
@@ -0,0 +1,1419 @@
...
+  cat("\nSelected:")
+
+  if (length(selected) == 0) {
+    cat(" (Empty)\n")
+  } else {
+    if (length(selected) == 1 && isRootTag(selected[[1]])) {
+      cat(" (Root)\n")
+    } else {
+      cat("\n")
+      print(tagQuerySelectedAsTags(selected))
```
cpsievert, line 790: Add a separator between selections (e.g., `------------------------`)

R/tag_env.R
wch: I think there are some times where one would legitimately want to allow empty selections. For example, suppose you want to write a function that finds tags with class `foo` and adds the class `bar`. You'd do something like:

```R
ensure_foobar <- function(x) {
  tagQuery(x)$find('.foo')$addClass('bar')$asTags()
}
```

But you wouldn't want this to throw a warning if the `.foo` selection was empty.

It might make sense to add an argument that tells it to warn on empty selections, but I think that's something we should come back to later.

schloerke: I think a dev mode warning is fair if you have not called `$find(".foo")` but are trying to call `$addClass("bar")` right away. 

I'm perfectly fine to not warn on a (user found) empty selection.

schloerke: Such as "Please call `$find()` before calling `$addClass()`

wch: I mean that it would be a general purpose function whose purpose is to ensure that the `bar` class is there along with `foo`, but the input may or may not have `foo`. Maybe this will illustrate better: imagine that you write a utility function that finds all `b` tags and replaces them with `i`, or one that capitalizes all text in `a()` tags, or something like that. Functions like that those should not throw warnings. Here's an analogy: imagine there was a global option to make `lapply` emit a warning if the input is length 0. That would be weird and emit unnecessary (and confusing) warnings.

schloerke: We now select all of the top level items by default. This avoids confusion all around.

R/selector.R
```diff
@@ -0,0 +1,279 @@
...
+}
+#' @export
+print.htmltools.selector.list <- function(x, ...) {
+  cat("// htmltools css selector list\n")
+  cat(format(x, ...), "\n")
+}
+
+
+
+str_replace <- function(text, pattern, replacement, fixed = FALSE) {
```
wch, line 214: I don't think it makes sense to add the `fixed` parameter, since it doesn't exist in `stringr::str_replace()` -- if you want to use a parameter that's in `sub()`, why not just use `sub()`?

Same for `str_replace_all` below.

R/selector.R
```diff
@@ -0,0 +1,279 @@
...
+  if (length(first) == 1 && first == -1) {
+    return(NULL)
+  }
+
+  regmatches(x, regInfo)[[1]]
+}
+
+
+
+is_html <- function(x) {
```
wch, line 277: Is this functions used anywhere?

R/tag_query.R
```diff
@@ -0,0 +1,1601 @@
...
+  list(
+    push = stack$push,
+    asList = stack$as_list,
+    uniqueList = function() {
+      unique(stack$as_list())
+    }
+  )
+}
+
+# (Used for `unique_envirStack()` only. Do not use directly!)
```
wch, line 115:
```suggestion
# (Used for `unique_envirStack()` only. Do not use directly!)
# Provides same interface as `envirStack()`, but checks for duplicates when
# when items are on their way in (with `push()`) instead of on the way out
# (with `uniqueList()`). This is faster when size is ~500 and above.
```

R/tag_query.R
```diff
@@ -0,0 +1,1601 @@
...
+        # mark the key as _seen_
+        map$set(key, TRUE)
+        # add the env
+        stack$push(env)
+      }
+    },
+    uniqueList = stack$as_list
+  )
+}
+# Use to retrieve unique environments (eg: `tq$parent()`)
```
wch, line 132:
```suggestion
# Use to retrieve unique environments (eg: `tq$parent()`)
# Provides same interface as `envirStack()`, but switches to the faster
# `envirStackUni_()` implementation when size hits 500.
```

R/tag_query.R
```diff
@@ -0,0 +1,1601 @@
...
+  xList <- x
+
+  ret <- list2env(xList, new.env(parent = emptyenv()))
+
+  attrVals <- safeAttrValues(xList)
+  walk2(names(attrVals), attrVals, function(attrName, attrValue) {
+    attr(ret, attrName) <<- attrValue
+  })
+
+  class(ret) <- c(newClass, class(xList), "environment")
```
wch, line 179: Calling `class(xList)` on a plain list will return `"list"`, so the result may have the class `c("list", "environment")`.

I think that using `oldClass()` makes sense here, though I have to admit it's not 100% clear to me exactly what that function is supposed to do.

wch: From https://stat.ethz.ch/pipermail/r-help/2005-August/076881.html:

> so class returns the implicit class, and oldClass does not.

wch: Also, I'm curious: is the explicit `"environment"` class needed?

R/tag_query.R
```diff
@@ -0,0 +1,1601 @@
...
+  asTagEnv_(x, parent = x$parent)
+}
+asTagEnv_ <- function(x, parent = NULL, seenMap = envirMap()) {
+  isTagVal <- isTag(x)
+  isTagEnvVal <- isTagEnv(x)
+
+  if (isTagVal || isTagEnvVal) {
+    if (!isTagEnvVal) {
+      xList <- x
+      x <- safeListToEnv(xList, "htmltools.tag.env")
```
wch, line 219: Is there a particular reason not to hard-code `"htmltools.tag.env"` inside of `safeListToEnv()`?

R/tag_query.R
```diff
@@ -0,0 +1,1601 @@
...
+  map <- fastmap()
+  list(
+    keys = map$keys,
+    asList = function() {
+      unname(map$as_list())
+    },
+    has = function(envir) {
+      map$has(envir$envKey)
+    },
+    set = function(envir, value = envir) {
```
wch, line 95: This could be changed to:

```suggestion
    add = function(envir) {
      map$set(envir$envKey, TRUE)
```

right? Though I suppose the name `envirMap` should also be changed then, to something without "map" in it.

R/tag_query.R
```diff
@@ -0,0 +1,1601 @@
...
+        # Not necessarily the order of the circular dependency
+        # TODO-later show actual circular dependency and not all visited nodes? This should be rare
+        "Tags processed: (unordered set)\n", paste0("* ", seenMap$keys(), collapse = "\n")
+      )
+    }
+    # Add the item to the seen map to help with cycle detection
+    seenMap$set(x, TRUE)
+
+    # Make sure all attribs are unique
+    x$attribs <- flattenTagAttribs(x$attribs)
```
wch, line 236: We should be able to get rid of this once we implement flattening attributes on the way in.

R/tag_query.R
```diff
@@ -0,0 +1,1601 @@
...
+
+    # Recurse through children
+    if (length(x$children) != 0) {
+      # Possible optimization... name the children tags to the formatted values.
+      # * Allows for faster child look up later.
+      # * Comes with the cost of always formatting the env values even if children names are not needed.
+      # Attributes may be dropped
+      # * Could replace with `x$children[] <- ....`
+      # * Leaving as is to see if people mis-use the children field
+      x$children <- unname(lapply(
```
wch, line 246: Just curious: does it ever happen that `$children` is a named list?

R/tags.R
```diff
@@ -958,6 +954,49 @@ flattenTags <- function(x) {
     flattenTags(as.tags(x))
   }
 }
+# This method should be just like `flattenTags()`, except the final `else` will return `list(x)`, rather than calling `flattenTags(as.tags(x))`.
```
wch, line 957: Can you wrap all comments at 80 chars, like the RStudio IDE does it?

R/tags.R
```diff
@@ -958,6 +954,49 @@ flattenTags <- function(x) {
...
+# By not calling `as.tags(x)`, tagFunctions are not evaluated and other items are not converted.
+flattenTagsRaw <- function(x) {
```
wch, line 959: It's kind of unclear to me exactly what this function is supposed to do. It's called `flattenTagsRaw`, but it appears to wrap tags, tag-envs, and strings in lists, and it recurses into to lists/taglists.

So, this is like `flattenTags()`, except it accepts tag-envs, and it doesn't invoke tagFunctions?

R/tag_query.R
```diff
@@ -0,0 +1,1601 @@
...
+        # Does NOT recurse to grand-children etc.
+        flattenTagsRaw(x$children),
+        # recurse through each child
+        asTagEnv_,
+        parent = x,
+        seenMap = seenMap
+      ))
+    }
+    # Remove the item from the map to allow for checks for ciruclar deps
+    # Having multiple child objects that are the same is ok, as long as a cycle is not found
```
wch, line 257: When would it happen that multiple children would be the same? It's not clear to me why the `$remove()` is helpful.

R/tag_query.R
```diff
@@ -0,0 +1,1601 @@
...
+    seenMap$remove(x)
+  }
+  x
+}
+
+# This method MUST undo everything done in `asTagEnv(x)`
+# Do not export to encourage direct use of `tagQuery()$asTags()`
+tagEnvToTags <- function(x) {
+  if (isTagEnv(x)) {
```
wch, line 266: Maybe this should error if `x` is not a tag-env?

R/tag_query.R
```diff
@@ -0,0 +1,1601 @@
...
+    xEl <- x
+    # convert to list first to avoid altering the original env obj
+    x <- safeEnvToList(xEl, c("htmltools.tag.env"))
+    # undo parent env and key
+    x$parent <- NULL
+    x$envKey <- NULL
+    # recurse through children
+    if (!is.null(x$children)) {
```
wch, line 274: Is `x$children` ever NULL or not present?

R/tag_query.R
```diff
@@ -0,0 +1,1601 @@
...
+      x$children <- unname(lapply(x$children, tagEnvToTags))
```
wch, line 275: Same question as above about `$children` ever being named.

R/tag_query.R
```diff
@@ -0,0 +1,1601 @@
...
+#'   the single selected item. If any selections are made, the selected elements
+#'   will be updated.
+#' @md
+#' @export
+tagQuery <- function(tags) {
+
+  tags <- flattenTagsRaw(tags %||% list())
+  tags <- asTagEnv(
+    wrapWithRootTag(tags)
+  )
```
wch, line 440: So suppose the following is run:

```R
y <- tagQuery(x).find("div")
z <- tagQuery(y$get(1))
```

This is a fairly common operation with jQuery. Does this mean that the whole tree will take a round trip to a lists and back to environments?

And then modifying `z` won't affect `y`?

I think it may make more sense to keep it as environments and allow modifications to `z` to affect `y`, but we need to think through it to make sure this doesn't have weird consequences.

R/tag_query.R
```diff
@@ -0,0 +1,1601 @@
...
+        #' ## Tag Query methods
+        #'
+        #' * `$asTags(selected = TRUE)`: If `selected = TRUE`, then all
+        #' previously found elements (and their descendants) will be
+        #' converted to tags. If `selected = FALSE`, the top level tag
+        #' elements (and their descendants) will be converted to
+        #' standard tags. If there is more than one element being
+        #' returned, a `tagList()` will be used to hold all of the
+        #' objects.
+        asTags = function(selected = TRUE) {
```
wch, line 767: It's unfortunate that the return type of this isn't stable, but I can't think of a better way to go about it.

R/tag_query.R
```diff
@@ -0,0 +1,1601 @@
...
+  tags <- asTagEnv(
+    wrapWithRootTag(tags)
+  )
+
+  selected <- tagQueryFindReset(tags)
+
+  tagQuery_(tags, selected)
+}
+
+#' @rdname tagQuery
```
wch, line 447: I think this may be better if it's not publicly documented.

R/tag_query.R
wch: Another example:

```R
y <- tagQuery(x)
z <- tagQuery(y)
z$find("div")$addClass("foo")
```

I think it makes sense for this to modify `y` as well.

R/tag_query.R
```diff
@@ -0,0 +1,1601 @@
...
+
+  class(ret) <- setdiff(class(x), c(oldClass, "environment"))
+  ret
+}
+
+
+# Do not export to encourage direct use of `tagQuery()`
+asTagEnv <- function(x) {
+  if (isTagQuery(x)) {
+    return(asTagEnv(x$root()))
```
wch, line 201: `x$root()` will always be a tag env, right? If so, no need for `asTagEnv()`.

R/tag_query.R
```diff
@@ -0,0 +1,1601 @@
...
+  }
+
+  if (!isTagEnv(x)) {
+    if (!isTag(x)) {
```
wch, line 205: Any particular reason for this if-if structure instead of `if (a && b)`?

R/tag_query.R
wch: And now that I think of it, does this even need to accept tagQuery objects? If we don't actually use that anywhere, it would be conceptually simpler to just throw an error if a tagQuery object is passed in.

R/tag_query.R
```diff
@@ -0,0 +1,1601 @@
...
+  walk2(names(attrVals), attrVals, function(attrName, attrValue) {
+    attr(ret, attrName) <<- attrValue
+  })
+
+  class(ret) <- setdiff(class(x), c(oldClass, "environment"))
+  ret
+}
+
+
+# Do not export to encourage direct use of `tagQuery()`
```
wch, line 198: It would help to have a description of what this function is supposed to do here.

R/tag_query.R
schloerke: `oldClass(x)`... TIL

schloerke: > is the explicit `"environment"` class needed?

Not necessarily. But Hadley liked to add the class `"function"` to things like `shiny::reactive()` where it isn't necessary?  I was following his pattern.

R/tag_query.R
schloerke: I was trying to keep `safeListToEnv()` a generic function. Like something that `staticimports` could use.

It also makes it more explicit as to what is happening.

R/tag_query.R
schloerke: ... This is the way in?

R/tag_query.R
schloerke: I originally did this because `tag{Append,Set}Children()` call this internally. 

However, they do this as they are processing user input.  Removing...

R/tag_query.R
```diff
@@ -0,0 +1,1601 @@
...
+    stop("`tagQuery_(root=)` must be a tag environment")
+  }
+
+  # Use `var_` names to avoid namespace collision
+  # Make sure all elements are tag envs
+  rebuild_ <- function() {
+    # safe to do as `root` will never be turned into a standard list
+    asTagEnv(root)
+  }
+  selected_ <- selected
```
wch, line 462: I think it would be slightly cleaner to just take `selected_` as an argument to the `tagQuery_` function. Then if `selected_` is changed, we don't hang on to a reference of the original object, and there's also no confusion about the `selected` argument used in the internal functions.

R/tag_query.R
```diff
@@ -0,0 +1,1601 @@
...
+    selected <- selected %||% list()
+    if (!is.list(selected)) {
+      stop("`selected` must be a `list()`")
+    }
+    if (filterRoot) {
+      selected <- Filter(selected, f = function(s) {
+        isTagEnv(s) && !isRootTag(s)
+      })
+    } else {
+      selected <- Filter(selected, f = isTagEnv)
```
wch, line 481: Would the input `selected` ever have non-tag-env objects in it? If not, maybe this should error in case it encounters such an object?

R/tag_query.R
schloerke: It is safe to always call the `lapply`. Removing the check

R/tag_query.R
```diff
@@ -0,0 +1,1601 @@
...
+      filterStack$push(el)
+    }
+  })
+
+  filterStack$asList()
+}
+
+
+
+cssSelectorToSelector <- function(cssSelector) {
```
wch, line 1365: It would help to have a short explanation of what this function is supposed to do.

R/tag_query.R
```diff
@@ -0,0 +1,1601 @@
...
+tagQueryFindFilter <- function(els, fn) {
+  if (is.character(fn)) {
+    selector <- cssSelectorToSelector(fn)
+    fn <- function(el, i) {
+      elMatchesSelector(el, selector)
+    }
+  }
+  validateFnCanIterate(fn)
+
+  filterStack <- envirStack()
```
wch, line 1353: Should this be `envirStackUnique()`?

R/tag_query.R
```diff
@@ -0,0 +1,1601 @@
...
+  # For every element...
+  tagQueryWalk(els, function(el) {
+    if (!isTagEnv(el)) return()
+    # Ignore the element and
+    # Walk through each child...
+    tagQueryWalk(el$children, function(child) {
+      # Find descendant matching the `selector`
+      tagQueryFindDescendants_(child, selector, function(foundEl) {
+        foundStack$push(foundEl)
+      })
```
wch, line 1459:
```suggestion
      tagQueryFindDescendants_(child, selector, foundStack$push)
```

R/tag_query.R
schloerke: That `tagQuery_()` is not documented? or `tagQuery()`? Or both.

Currently there is no mention of `tagQuery_()` by setting `@aliases NULL` and `@usage NULL`.

R/tag_query.R
```diff
@@ -0,0 +1,1601 @@
...
+    })
+  })
+  siblingStack$uniqueList()
+}
+
+# Filter the selected elements using a function
+# The answer of `fn(el, i)` should work in an `if` block
+tagQueryFindFilter <- function(els, fn) {
+  if (is.character(fn)) {
+    selector <- cssSelectorToSelector(fn)
```
wch, line 1346: It would help to clarify the difference between a `cssSelector` and a `selector`. (Maybe not here, but somewhere)

R/tag_query.R
schloerke: We could always return a `tagList()`, even if there is a single element. 

`flattenTagsRaw()` will make sure it doesn't ever wrap `tagList()`s on `tagList()`s.

R/tag_query.R
schloerke: I am going to leave it as `envirMap`. If it was just `map`, then I agree that the name should be changed.

R/tag_query.R
schloerke: (leaving as is)

R/tag_query.R
schloerke: Yes, we can have y & z share the same tag envs.

--------------------------

Let's say we do 

```r
y <- tagQuery(x)$find(".foo")
z <- tagQuery(y)
```

`y` now has selected elements within the tag environment of `x`. 

Question:
* Should `z` be the root of `y` with no selected elements? (Current implementation; Like passing DOM to another `tagQuery()`)
* Or should `z` be the root of `y` with the same selected elements of `y`? (Similar to jQuery;)

wch: If `tagQuery()` is called on a tag-env object, then traverse up all the parents until you hit the root, and build a new tagQuery object with that root, and the tag-env as the selection.

schloerke: Implemented in 28183fd

R/tag_query.R
schloerke: Since all of the others are unique stacks, I'll make this one too. 

(I was operating under the assumption that the selected elements would already be unique. So it would be ok to just use a regular stack to filter it. Can't get any _worse_)

schloerke: e0a581d

R/tag_query.R
schloerke: 8242d96

R/tags.R
schloerke: 👍

R/tag_query.R
wch: After discussion, we decided to remove it.

R/tag_query.R
```diff
@@ -0,0 +1,1752 @@
...
+safeEnvToList <- function(x, oldClass = NULL) {
+  xEnv <- x
+  ret <- as.list.environment(xEnv, all.names = TRUE)
+
+  attrVals <- safeAttrValues(xEnv)
+  walk2(names(attrVals), attrVals, function(attrName, attrValue) {
+    attr(ret, attrName) <<- attrValue
+  })
+
+  class(ret) <- setdiff(class(x), c(oldClass, "environment"))
```
wch, line 219: Same as above regarding `oldClass` and `"environment"`.

R/tag_query.R
wch: See #205.

R/tag_query.R
```diff
@@ -0,0 +1,1752 @@
...
+        flattenTagsRaw(x$children),
+        # recurse through each child
+        asTagEnv_,
+        parent = x,
+        seenMap = seenMap
+      )
+    }
+    # Remove the item from the map to allow for checks for ciruclar deps
+    # Having multiple child objects that are the same is ok, as long as a cycle is not found
+    seenMap$remove(x)
```
wch, line 296: Remove this and do this after a cycle is detected.

R/tag_query.R
```diff
@@ -0,0 +1,1752 @@
...
+      # add parent env and key
+      x$parent <- parent
+      x$envKey <- sexp_address(x)
+    }
+    if (seenMap$has(x)) {
+      stop(
+        "Circular family tree found with tag environment: ", sexp_address(x), "\n",
+        # Not necessarily the order of the circular dependency
+        # TODO-later show actual circular dependency and not all visited nodes? This should be rare
+        "Tags processed: (unordered set)\n", paste0("* ", seenMap$keys(), collapse = "\n")
```
wch, line 267: Add message about using `lobstr::tree(show_environments = TRUE)` to see the cycle.

R/tag_query.R
```diff
@@ -0,0 +1,1752 @@
...
+
+# This method MUST undo everything done in `asTagEnv(x)`
+# Do not export to encourage direct use of `tagQuery()$asTags()`
+tagEnvToTags <- function(x) {
+  if (!isTagEnv(x)) {
+    stop("`tagEnvToTags(x)` must start with a tag environment")
+  }
+  tagEnvToTags_(x)
+}
+tagEnvToTags_ <- function(x) {
```
wch, line 309: Can you add a short comment about the types that this can accept (due to recursion passing in other stuff).

R/tag_query.R
```diff
@@ -0,0 +1,1752 @@
...
+      if (!isTagEnv(el)) return(FALSE)
+      tagHasAttribute(el, attr)
+    }),
+    use.names = FALSE
+  )
+}
+
+
+getCssClass <- function(class) {
+  class <- unlist(list2(class))
```
wch, line 1287:
```suggestion
  class <- as.character(list2(class))
```

Here and elsewhere that `unlist()` is used.

R/tag_query.R
```diff
@@ -0,0 +1,1752 @@
+#' @importFrom fastmap fastmap faststack
```
wch, line 1: Should make a pkgdown site.

wch (approved):

R/tag_query.R
```diff
@@ -0,0 +1,1752 @@
...
+  parentStack$uniqueList()
+}
+# Return a list of the unique set of ancestor elements
+
+# * By only looking for elements that have not been seen before, searching is as
+# lazy as possible
+# * Must traverse all parents; If cssSelector exists, only return found parents
+# that match selector.
+# * Search using breadth-first. This is as close to jQuery's implementation.
+# * (I'd prefer to do depth first, like `$closest()`. No need to make a new
```
wch, line 1389: Change implementation to depth-first.

R/tag_query.R
schloerke: Going to not allow duplicate tag environments. If the user accidentally puts the same child in two separate locations, then the parent pointer is invalid for one of the locations.

R/tag_query.R
schloerke: Will add in another PR. (I think this will be breaking behavior)

R/tags.R
schloerke: Skipping for now

R/tag_query.R
schloerke: Will do in a followup PR

PR #212, opened 2021-04-15
`tagGetAttribute(x, attr)` returns a list if one of the values is not atomic

### LINKED ISSUE #205: Tag classes should be normalized on creation

schloerke:
Could not approach by flattening and splitting values. Too many downstream dependencies reach into the `attribs`.

Fixes #205 

To get around more complicated attribs values, a list will be returned if any value is not atomic. We still recommend that everyone use character values only (or values that can be pasted together).


R/tags.R
```diff
@@ -309,19 +311,10 @@ tagHasAttribute <- function(tag, attr) {
...
-
-  # Convert all attribs to chars explicitly; prevents us from messing up factors
-  result <- lapply(attribs[attrIdx], as.character)
-  # Separate multiple attributes with the same name
-  result <- paste(result, collapse  = " ")
-  result
+  # Must flatten all tag attribs to handle legacy situations
+  #   where there are two entries for the same name
+  attribs <- flattenTagAttribs(tag$attribs)
+  attribs[[attr]]
```
cpsievert, line 317: This'll end up calling `as.character()` on _all_ the attributes now, right? Seems like we should avoid doing that (I think using `as.character()` here  was a mistake in the 1st place)

R/tags.R
```diff
@@ -412,7 +405,9 @@ tag <- function(`_tag_name`, varArgs, .noWS=NULL) {
 
   # Named arguments become attribs, dropping NULL and length-0 values
   named_idx <- nzchar(varArgsNames)
-  attribs <- dropNullsOrEmpty(varArgs[named_idx])
+  attribs <- flattenTagAttribs(
+    dropNullsOrEmpty(varArgs[named_idx])
+  )
```
cpsievert, line 410: I think we'll want to avoid `as.character` here as well.

Also, in the current implementation of `flattenTagAttribs` seems we could get a speed improvement by using `anyNA(x)` instead of `any(is.na(x))`

R/tags.R
```diff
@@ -292,7 +292,9 @@ tagFunction <- function(func) {
 #' @export
 tagAppendAttributes <- function(tag, ...) {
   throw_if_tag_function(tag)
-  tag$attribs <- c(tag$attribs, dropNullsOrEmpty(dots_list(...)))
+  tag$attribs <- flattenTagAttribs(
+    c(tag$attribs, dropNullsOrEmpty(dots_list(...)))
+  )
```
cpsievert, line 297: Same as other comments (avoid `as.character()` until print-time)

NEWS.md
```diff
@@ -1,5 +1,9 @@
 # htmltools (development version)
 
+## Breaking Changes
+
+* Tag attributes are now combined when the tag object is created, when appending new tag attributes, and when retrieving tag attributes. For example, before this change, it was possible to have attributes that looked like `list(class = "A", class = "B")`. Now, it will be stored as `list(class = "A B")`. (#212)
```
wch, line 5:
```suggestion
* Closed #205: Tag attributes are now combined when the tag object is created, when appending new tag attributes, and when retrieving tag attributes. For example, before this change, it was possible to have attributes that looked like `list(class = "A", class = "B")`. Now, it will be stored as `list(class = "A B")`. (#212)
```

tests/testthat/test-tags.r
```diff
@@ -893,3 +893,19 @@ test_that("extractPreserveChunks works for emoji strings", {
     c('chunk2', 'chunk1')
   )
 })
+
+
+test_that("tag attributes are normalized on creation", {
+  d <- div(class = "a", "text1", class = "b c", "text2")
```
wch, line 899: Can you also test when a vector is used, like `c("d", "e")`? See #205.

R/tags.R
schloerke: Should we take the stance that all values should be kept in an array?

So if it is a character and if we detect a space, then split the value into parts.  Otherwise, only collect the values.

Then at print time, we can convert to character and paste together.

wch: Yeah, I think it does make sense to always store `"a b"` as `c("a", "b")`. However, we have to be careful about handling other types -- for example, lists, factors, and even tag objects.

See https://github.com/rstudio/shiny/pull/3353 for an example of where this was an issue.

Maybe we need one code path for handling the common case, where all instances of a particular attribute are character vectors (as in `div(a = "x", a = "y")`).

But if any instances of an attribute are a different type, just leave them untouched. It might make sense to also warn if there are multiple instances of an attribute where at least one is not a char vector, as in `div(a = "x", a = span("hello"))`.

wch:
Following up a discussion we had: We decided that we would always collapse multiple entries with the same name into a single vector/list, allowing type promotion to happen normally.

For example, `div(x = "a b", x = "c")` would result in an `attribs` data structure of `c("a b", "c")`.

If the input includes a list or other complex data structure, like `div(x = c("a", "b"), x = span("hello"))`, the `attribs` data structure would be `list(c("a", "b"), span("hello"))`. (Note that this would generally be a bad idea to use!)

Note that we also need to be a bit careful about factors.

An additional thought I had afterward: we could add a parameter named `split` to `tagGetAttribute()` to do the following:

```R
z <- div(x = "a b", x = "c")

tagGetAttribute(z, "x")
#> [1] "a b c"

tagGetAttribute(z, "x", split = TRUE)
#> [1] "a" "b" "c"
```

This would make it more convenient to check if the attribute contains a specific value, like `"b"`.

I'm not sure the best way to handle attributes which are lists here, though.

schloerke:
With the latest implementation...

```r
z <- div(a = "x", a = "y")

z
#> <div a="x y"></div>

tagGetAttribute(z, "a")
#> [1] "x y"
```

---------------------

```r
w <- div(a = "x", a = list(obj = TRUE))

tagGetAttribute(w, "a")
#> [[1]]
#> [1] "x"
#> 
#> [[2]]
#> [[2]]$obj
#> [1] TRUE

w
#> <div a="x TRUE"></div>
```

R/tags.R
schloerke: See https://github.com/rstudio/htmltools/pull/212#issuecomment-823559692

tests/testthat/test-tags.r
schloerke: a0dfdeb

R/tag_query.R
```diff
@@ -1239,9 +1235,9 @@ tagQueryAttrsRemove <- function(els, attrs) {
   }
   tagQueryWalk(els, function(el) {
     if (!isTagEnv(el)) return()
-    for (attrVal in attrs) {
-      el$attribs[[attrVal]] <- NULL
-    }
+    # Alter in place
+    # Remove locations that have a matching name
+    el$attribs[names(el$attribs) %in% attrs] <- NULL
```
schloerke, line 1240:
```suggestion
    el$attribs[names2(el$attribs) %in% attrs] <- NULL
```

scripts/reprex/flattenTagAttribs/flattenTagAttribs.Rmd
```diff
@@ -0,0 +1,111 @@
+---
+title: "Stack benchmarks"
```
schloerke, line 2:
```suggestion
title: "flattenTagAttribs benchmarks"
```

schloerke:
I think this PR is ready to be merged. 

The [comment above](https://github.com/rstudio/htmltools/pull/212#issuecomment-823559692) shows a typical use case and a not-recommended use case.

The internal `attrib` structures are not altered and are only adjusted when printed or retrieved with `tagGetAttribute()`

R/tags.R
```diff
@@ -427,10 +427,24 @@ tagGetAttribute <- function(tag, attr) {
...
-  # Convert all attribs to chars explicitly; prevents us from messing up factors
-  result <- lapply(attribs[attrIdx], as.character)
-  # Separate multiple attributes with the same name
-  result <- paste(result, collapse  = " ")
+  result <- attribs[attrIdx]
+  if (all(vapply(result, is.atomic, logical(1)))) {
+    # Convert all attribs to chars explicitly; prevents us from messing up factors
+    # Separate multiple attributes with the same name
+    vals <- lapply(result, function(val) {
+      val <- as.character(val)
```
cpsievert, line 435: One thing that has annoyed me about how `tagGetAttribute()` currently works is this behavior:

```r
> div(foo=NA)
<div foo></div>
> tagGetAttribute(div(foo=NA), "foo")
[1] "NA"
```

Maybe it'd make since to _not_ convert to character when `isTRUE(is.na(val))`?

schloerke: Absolutely! Great catch

schloerke: Using the cran version...

```r
tagGetAttribute(div(foo=NA), "foo")
[1] "NA"
```

I still believe the change is correct.

schloerke: Addressed in 26b3042

NEWS.md
```diff
@@ -8,6 +12,8 @@
 
 ## Bug Fixes
 
+* When retrieving a tag attribute using `tagGetAttribute(tag, attr)`, `NA` values will be removed before combing remaining attribute values. If all attribute values are `NA`, then a single `NA` value will be returned. (#212)
```
cpsievert, line 15:
```suggestion
* When retrieving a tag attribute using `tagGetAttribute(tag, attr)`, `NA` values will be removed before combining remaining attribute values. If all attribute values are `NA`, then a single `NA` value will be returned. (#212)
```

R/tags.R
```diff
@@ -427,10 +427,33 @@ tagGetAttribute <- function(tag, attr) {
...
+    if (all(na_idx)) {
+      return(NA)
+    }
+    result <- result[!na_idx]
+  }
+
+  if (all(vapply(result, is.atomic, logical(1)))) {
+    # Convert all attribs to chars explicitly; prevents us from messing up factors
+    # Separate multiple attributes with the same name
+    vals <- lapply(result, function(val) {
```
cpsievert, line 443: I suppose this could be a `vapply()`?

cpsievert (approved):
LGTM pending minor suggestions

R/tags.R
schloerke: 10dd52b

cpsievert (approved):

PR #214, opened 2021-04-15
Display the raw `tagname` in the tag functions

schloerke:
```r
div
#> function (..., .noWS = NULL, .render = NULL)
#> {
#>     validateNoWS(.noWS)
#>     contents <- dots_list(...)
#>     tag("div", contents, .noWS = .noWS, .render = .render)
#> }
```

### DIFF
--- a/R/tags.R
+++ b/R/tags.R
@@ -843,15 +843,16 @@ names(known_tags) <- known_tags
 #' @keywords NULL
 #' @import rlang
 tags <- lapply(known_tags, function(tagname) {
-  # Make the function signature
-  tagFunc <- function(..., .noWS = NULL, .render = NULL) {}
   # Overwrite the body with the `tagname` value injected into the body
-  rlang::fn_body(tagFunc) <- rlang::inject({rlang::expr({
-    validateNoWS(.noWS)
-    contents <- dots_list(...)
-    tag(!!tagname, contents, .noWS=.noWS, .render = .render)
-  })})
-  tagFunc
+  new_function(
+    args = exprs(... = , .noWS = NULL),
+    expr({
+      validateNoWS(.noWS)
+      contents <- dots_list(...)
+      tag(!!tagname, contents, .noWS=.noWS)
+    }),
+    env = asNamespace("htmltools")
+  )
 })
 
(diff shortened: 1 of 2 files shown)


schloerke:
I've changed the base branch to https://github.com/rstudio/htmltools/pull/215 as this method will need to be adjusted.

The GHA will not run anymore as the branch is not master.

If the concept is good, we can merge it and have https://github.com/rstudio/htmltools/pull/215 run the checks.

R/tags.R
```diff
@@ -843,11 +843,15 @@ names(known_tags) <- known_tags
...
+  tagFunc <- function(..., .noWS = NULL, .render = NULL) {}
+  # Overwrite the body with the `tagname` value injected into the body
+  rlang::fn_body(tagFunc) <- rlang::inject({rlang::expr({
     validateNoWS(.noWS)
     contents <- dots_list(...)
-    tag(tagname, contents, .noWS=.noWS)
-  }
+    tag(!!tagname, contents, .noWS=.noWS, .render = .render)
+  })})
+  tagFunc
```
cpsievert, line 854: Would it make sense to use `rlang::new_function`?

schloerke: I could not get the arguments to include `...`. (Doesn't mean it can't be done. Just didn't figure it out.)

The body argument needs to be the result of the `inject` call.

cpsievert (approved):

R/tags.R
```diff
@@ -843,11 +843,16 @@ names(known_tags) <- known_tags
...
-  }
+  # Overwrite the body with the `tagname` value injected into the body
+  new_function(
+    args = exprs(... = , .noWS = NULL),
+    expr({
+      validateNoWS(.noWS)
+      contents <- dots_list(...)
+      tag(!!tagname, contents, .noWS=.noWS)
+    }),
+    env = asNamespace("htmltools")
```
wch, line 854: Does the namespace exist when this is run? Note that this may differ between `load_all()` and a real package building scenario.

wch: Note: I just tested and `asNamespace("htmltools")` does seem to work during package build.

wch (approved):

PR #215, opened 2021-04-15
Add `tagAddRenderHook()` function

schloerke:
`tagFunction()`s are black boxes and do not provide plausible tag structures.  Typically with `tagFunction()`s, the tag structure is known and only a class value or html dependency is added.

To provide a better default tag structure, `TAG$.renderHooks` will contain a list of render functions. When converted to tags with `as.tags(TAG)`, if a render hook exists, `as.tags.shiny.tag(TAG)` will remove a single render hook, call it, and call `as.tags(TAG2)`. This recursion will continue until no more render hooks exist. (To be safe, the hook is removed from `TAG` before calling `hook(TAG)`.)

Using `tagAddRenderHook()` (mixed with `tagQuery()`) will be preferred over using generic `tagFunction()`s.


R/tags.R
```diff
@@ -921,10 +1031,26 @@ withTags <- function(code) {
...
+      if (is.function(renderFn)) {
+        # Set render method to `NULL` to avoid confusion
+        uiObj$.render <- NULL
+        return(
+          tagify(renderFn(uiObj))
+        )
+      } else {
+        return(uiObj)
+      }
+    }
```
cpsievert, line 1045: I think this should happen in an `as.tags` method (similar to `as.tags.shiny.tag.function`)? That way if we want to add `.render` to `tagList()`, doing so will be simpler (and we can keep the implementation of `tagify()` 'simple')

R/tags.R
```diff
@@ -383,6 +474,9 @@ throw_if_tag_function <- function(tag) {
 #'   normally be written around this tag. Valid options include `before`,
 #'   `after`, `outside`, `after-begin`, and `before-end`.
 #'   Any number of these options can be specified.
+#' @param .render A function that is called during render time. This function
```
cpsievert, line 477: Part of me wanted this to be `.onRender` because of the precedence in `htmlwidgets::createWidget()`, but I don't feel super strongly about it

R/tags.R
```diff
@@ -288,6 +291,94 @@ tagFunction <- function(func) {
...
+  prevFunc <- tag$.render
+
+  if (!is.function(prevFunc) || isTRUE(replace)) {
+    tag$.render <- func
+    return(tag)
+  }
+
+  tag$.render <- function(x) {
+    func(prevFunc(x))
+  }
```
cpsievert, line 376: Nice!

tests/testthat/test-tags.r
```diff
@@ -893,3 +893,68 @@ test_that("extractPreserveChunks works for emoji strings", {
...
+
+  # Replace the previous render method
+  # Should print a `div` with class `"extra"`
+  divExtra <- tagAddRenderFunction(obj, replace = TRUE, function(x) {
+    tagAppendAttributes(x, class = "extra")
+  })
+  expect_equal(divExtra$attribs$class, NULL)
+  expect_equal(tagify(divExtra)$name, "div")
+  expect_equal(tagify(divExtra)$attribs$class, "extra")
+
```
wch, line 925: It would be good to also add a test where another layer of render function is added with `tagAddRenderFunction()`.

R/tags.R
```diff
@@ -288,6 +291,94 @@ tagFunction <- function(func) {
...
+
+  prevFunc <- tag$.render
+
+  if (!is.function(prevFunc) || isTRUE(replace)) {
+    tag$.render <- func
+    return(tag)
+  }
+
+  tag$.render <- function(x) {
+    func(prevFunc(x))
```
wch, line 375: Subtle issue here, which may be relevant. If I were a user, I'd expect the functions to be executed in the order that they're added, but because of R's lazy evaluation, this could cause them to execute in a different order, depending on the contents of the functions. This may be relevant in cases where the functions are altering and reading external state, which may be the case when (for example) setting the current theme.

For example, consider these functions:

```R
times_10 <- FALSE

first <- function(x) {
  times_10 <<- TRUE
  x
}

second <- function(x) {
  if (times_10) x * 10
  else              x
}
```

If you call `first()` and save the value, and then call `second()` on that value, you'd see this:

```R
times_10 <- FALSE
y <- first(1)
y <- second(y)
y
#> [1] 10
```

But if you call `second(first(1))`, you get a different answer:

```R
times_10 <- FALSE
y <- second(first(1))
y
#> [1] 1
```

So I think it would make sense to call the two functions separately, instead of nested.

wch: Oh, and I suppose there should be a test for this.

R/tags.R
```diff
@@ -288,6 +295,97 @@ tagFunction <- function(func) {
...
 }
 
+#' Add a tag render function
+#'
+#' Adds a render method to a tag object. This allows for the tag structure to
+#' be physically present while still allowing for a function to enhance (or
+#' even completely replace) the original tag object.
+#'
+#' It is recommended to use a render function over a [`tagFunction()`] whenever
+#' possible. By using a render method, a the tag structure is not a black box
```
cpsievert, line 305:
```suggestion
#' possible. By using a render method, the tag structure is not a black box
```

R/tags.R
```diff
@@ -288,6 +295,97 @@ tagFunction <- function(func) {
...
+#'   script = "jquery-ui.min.js")
+#' # At render time, both dependencies will be found
+#' renderTags(newDep)$dependencies
+#'
+#' # Ignore the original tag and return something completely new.
+#' newObj <- tagAddRender(obj, function(x) {
+#'   tags$p("Something else")
+#' })
+#' newObj
+tagAddRender <- function(tag, func, add = TRUE) {
```
cpsievert, line 367: Thinking about this again, maybe this could be `tagRenderHook()` and `.render` could be `.renderHook` (I was wrong about htmlwidgets using `onRender` -- it's actually `preRenderHook`)?

schloerke: Works for me!

cpsievert (approved):

NEWS.md
```diff
@@ -4,6 +4,8 @@
 
 * Added `tagQuery(tags)`. A tag query object implements many popular features of jQuery.  Similar to jQuery, tag query objects can find internal html using CSS selections. Given a selection (which defaults to the original `tags`), many alterations may be performed before converting the tag query object back to tag objects. (#208)
 
+* Added a new `tagRenderHook()` for delaying modification of a tag object until it is rendered. A render-time hook may also be added via the new `.renderHook` argument added to all `tag()` functions. (#215)
```
cpsievert, line 7:
```suggestion
* Added`tagRenderHook()` for delaying modification of a tag object until it is rendered. A render-time hook may also be added via the new `.renderHook` argument added to all `tag()` functions. (#215) 
```

wch:
The other `tag*` functions have a verb in the name, like `tagAddChildren()`, `tagAppendAttributes()`, etc. Can we rename this so that there's a verb in the name, as in `tagAddRenderHook`?

R/tags.R
```diff
@@ -288,6 +295,100 @@ tagFunction <- function(func) {
...
+#'   script = "jquery-ui.min.js")
+#' # At render time, both dependencies will be found
+#' renderTags(newDep)$dependencies
+#'
+#' # Ignore the original tag and return something completely new.
+#' newObj <- tagRenderHook(obj, function(x) {
+#'   tags$p("Something else")
+#' })
+#' newObj
+tagRenderHook <- function(tag, func, add = TRUE) {
```
wch, line 370:
```suggestion
tagAddRenderHook <- function(tag, func, replace = FALSE) {
```

R/tags.R
```diff
@@ -288,6 +295,100 @@ tagFunction <- function(func) {
...
+  }
+
+  prevFunc <- tag$.renderHook
+
+  if (!is.function(prevFunc) || !isTRUE(add)) {
+    tag$.renderHook <- func
+    return(tag)
+  }
+
+  tag$.renderHook <- function(x) {
```
wch, line 382: Instead of wrapping this stuff in closures, we should make `tag$.renderHook` a list, so that users are able to inspect render hook functions. (With closures, it's very difficult for a user to see how many render hooks there are, and what each one does.)

Then at render time, it can iterate through the list of functions, executing each one and passing along the returned value.

schloerke: similar to shiny route handlers?

R/tags.R
```diff
@@ -403,7 +502,18 @@ throw_if_tag_function <- function(tag) {
...
+#' h <- tags$h1("Example", .renderHooks = function(x) {
+#'   x$name <- "h2"
+#'   x
+#' })
+#' h
+#' oneline <- tag("span",
+#'   tag("strong", "Super strong", .noWS="outside")
+#' )
+#' cat(as.character(oneline))
+tag <- function(`_tag_name`, varArgs, .noWS = NULL, .renderHooks = NULL) {
```
cpsievert, line 516: This should be a singular `.renderHook`, right?

R/tags.R
```diff
@@ -383,6 +478,10 @@ throw_if_tag_function <- function(tag) {
 #'   normally be written around this tag. Valid options include `before`,
 #'   `after`, `outside`, `after-begin`, and `before-end`.
 #'   Any number of these options can be specified.
+#' @param .renderHooks A function to call when the `tag` is rendered. This
+#'   function should have at least one argument (the `tag`) and return anything
+#'   that can be converted into tags via [as.tags()]. Multiple hooks may also be
+#'   added to a particular `tag` via [tagAddRenderHook()].
```
cpsievert, line 484: Winston and I were thinking this would just allow a function, not a list of functions (if you really needed that you could write a function to call the list of functions).

R/tags.R
schloerke: If its a list of functions, I'd like to keep it as `.renderHooks`

R/tags.R
schloerke: https://github.com/rstudio/htmltools/pull/215#discussion_r617005722

Winston asked for a list of functions.

cpsievert: Yea, we'll always attach a list of functions to the tag object, but when receiving input it should be a singular `.renderHook`

wch: Just to clarify: the data structure for storage would be a list of functions, but this function would accept just a single function, for simplicity and consistency. I think it would be very rare for someone to want to pass in multiple render hook functions; if they really want to do that, they could call `tagAddRenderHook()` (possibly multiple times), or they could wrap their set of functions in their own closure which invokes them.

schloerke:
(The error above is a GHA false positive error due to a caching failure.)

NEWS.md
```diff
@@ -4,6 +4,8 @@
 
 * Added `tagQuery(tags)`. A tag query object implements many popular features of jQuery.  Similar to jQuery, tag query objects can find internal html using CSS selections. Given a selection (which defaults to the original `tags`), many alterations may be performed before converting the tag query object back to tag objects. (#208)
 
+* Added `tagAddRenderHook()` for delaying modification of a tag object until it is rendered. A list of render-time hooks may also be added via the new `.renderHooks` argument added to all `tag()` functions. (#215)
```
cpsievert, line 7:
```suggestion
* Added `tagAddRenderHook()` for delaying modification of a tag object until it is rendered. A list of render-time hooks may also be added via the new `.renderHook` argument added to all `tag()` functions. (#215)
```

R/tags.R
```diff
@@ -383,6 +478,10 @@ throw_if_tag_function <- function(tag) {
 #'   normally be written around this tag. Valid options include `before`,
 #'   `after`, `outside`, `after-begin`, and `before-end`.
 #'   Any number of these options can be specified.
+#' @param .renderHook A function (or list of functions) to call when the `tag` is rendered. This
+#'   function should have at least one argument (the `tag`) and return anything
+#'   that can be converted into tags via [as.tags()]. Multiple hooks may also be
```
cpsievert, line 483:
```suggestion
#'   that can be converted into tags via [as.tags()]. Additional hooks may also be
```

R/tags.R
```diff
@@ -921,10 +1045,13 @@ withTags <- function(code) {
...
-    if (isTag(uiObj) || isTagList(uiObj) || is.character(uiObj))
+    if (isResolvedTag(uiObj) || isTagList(uiObj) || is.character(uiObj)) {
       return(uiObj)
-    else
-      return(tagify(as.tags(uiObj)))
+    }
+
+    return(
+      tagify(as.tags(uiObj))
+    )
```
cpsievert, line 1054: Can we reduce the diff here for the sake of clean history?

cpsievert:
LGTM pending minor nitpicks 👍

Issue #221, opened 2021-04-20
Consider renaming `$asTags(selected = TRUE)` 


wch:
It's a bit strange that:
* To get the whole tagQuery object back as a tag object, you call `$asTags(selected=FALSE)`.
* To get just the selected elements as tag objects, you call `$asTags(selected=TRUE)`.

The naming is a bit counterintuitive.

Maybe we should use:
* `$asTags()`
* `$selectedAsTags()` or `$asTagsSelected()`

Or possibly:
* `$asTags(all = TRUE)` (this would be the default)
* `$asTags(all = FALSE)`

cpsievert:
Now that I'm writing about this, I almost want to get rid of `$asTags()` and instead have a `$root()` and `$selected()` them both return tags, does that seem reasonable @schloerke?

cpsievert:
(Also, maybe it makes sense to do this as a part of https://github.com/rstudio/htmltools/issues/213)

schloerke:
Two issues with one PR. I'm for it!

Issue #213, opened 2021-04-15
Remove tagQuery methods `$selected()`, `$root()`, and $get(position)`


schloerke:
Having access to the items gives more confusion then benefit. 

* `$root()` is not necessary as any other methods should be accomplished by using a `tagQuery()` object.
* `$selected()` list operations should be instead by completed by `$each(fn)`
* `$get(position)` just a faster form of `$each(fn)`

cc @cpsievert

PR #227, opened 2021-04-21
Have `$selected()` and `$root()` return tags. Remove `$asTags()` and `$get()`

schloerke:
Fixes https://github.com/rstudio/htmltools/issues/221
Fixes https://github.com/rstudio/htmltools/issues/213

### DIFF
--- a/R/tag_query.R
+++ b/R/tag_query.R
@@ -376,15 +376,15 @@ print.htmltools.tag.query <- function(x, ...) {
 #' @export
 format.htmltools.tag.query <- function(x, ...) {
   stop(
-    "`format.htmltools.tag.query(x)` not allowed.\n",
-    "Please call `format()` the result of `$selected()`"
+    "`tagQuery()` objects can not be written directly as HTML tags.\n",
+    "Call either `$root()` or `$selected()` to extract the tags of interest."
   )
 }
 #' @export
 as.character.htmltools.tag.query <- function(x, ...) {
   stop(
-    "`as.character.htmltools.tag.query(x)` not allowed.\n",
-    "Please call `as.character()` the result of `$selected()`"
+    "`tagQuery()` objects can not be written directly as HTML tags.\n",
+    "Call either `$root()` or `$selected()` to extract the tags of interest."
   )
 }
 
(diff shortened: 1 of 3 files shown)


wch:
After this change, is there a way to get the root or selected tag-env objects?

cpsievert:
> After this change, is there a way to get the root or selected tag-env objects?

If users really need this ability (do they?), perhaps we could have `$_root()` and `$_selected()` methods (or maybe add a `asTagEnv = FALSE` argument to each)?

schloerke:
They can access them using:

* Selected elements: `$each(func)` 
* Root elements: `$reset()$each(func)`

R/tag_query.R
```diff
@@ -377,14 +377,14 @@ print.htmltools.tag.query <- function(x, ...) {
 format.htmltools.tag.query <- function(x, ...) {
   stop(
     "`format.htmltools.tag.query(x)` not allowed.\n",
-    "Please call `format()` the result of `$asTags()`"
+    "Please call `format()` the result of `$selected()`"
```
cpsievert, line 380: "`tagQuery()` objects can not be written directly as HTML tags.\n", 
    "Call either `$root()` or `$selected()` to extract the tags of interest."

R/tag_query.R
```diff
@@ -377,14 +377,14 @@ print.htmltools.tag.query <- function(x, ...) {
...
   )
 }
 #' @export
 as.character.htmltools.tag.query <- function(x, ...) {
   stop(
     "`as.character.htmltools.tag.query(x)` not allowed.\n",
-    "Please call `as.character()` the result of `$asTags()`"
+    "Please call `as.character()` the result of `$selected()`"
```
cpsievert, line 387: "`tagQuery()` objects can not be written directly as HTML tags.\n", 
    "Call either `$root()` or `$selected()` to extract the tags of interest."

cpsievert (approved):
Thanks!

wch:
I don't think it would be commonly used to access them directly. I could see accessing tag-envs as being useful for debugging, and possibly for doing operations that don't yet have `tagQuery` convenience functions, but I don't know for sure. The `each(func)` method seems sufficient for that, I think.

Issue #222, opened 2021-04-20
$addAttrs() should throw a better error if no names are supplied


cpsievert:
```r
> (tagQuery(div())$addAttrs("a"))
Root:
Error in if (!is.na(attribValue)) { : argument is of length zero
```

PR #229, opened 2021-04-21
Provide better error message when adding unnamed attrib values

schloerke:
Fixes https://github.com/rstudio/htmltools/issues/222

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -8,7 +8,7 @@
 
 * Closed #197: Fixed rendering of boolean attributes in <script> tags rendered via renderDependencies() (#197, thanks @atusy).
 
-* Unnamed attributes are no longer allowed to be appended via `tagAppendAttribs()`. When trying to print unnamed tag attribs, a better error message is provided. (#229)
+* Closed #222; Unnamed attributes are no longer allowed to be appended via `tagAppendAttribs()`. When trying to print unnamed tag attribs, a better error message is provided. (#229)
 
 
 # htmltools 0.5.1.1
--- a/NEWS.md
+++ b/NEWS.md
@@ -8,6 +8,8 @@
 
 * Closed #197: Fixed rendering of boolean attributes in <script> tags rendered via renderDependencies() (#197, thanks @atusy).
 
+* Unnamed attributes are no longer allowed to be appended via `tagAppendAttribs()`. When trying to print unnamed tag attribs, a better error message is provided. (#229)
+
 
 # htmltools 0.5.1.1
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -8,7 +8,7 @@
 
 * Closed #197: Fixed rendering of boolean attributes in <script> tags rendered via renderDependencies() (#197, thanks @atusy).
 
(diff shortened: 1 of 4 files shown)


R/tags.R
```diff
@@ -292,7 +292,11 @@ tagFunction <- function(func) {
 #' @export
 tagAppendAttributes <- function(tag, ...) {
   throw_if_tag_function(tag)
-  tag$attribs <- c(tag$attribs, dropNullsOrEmpty(dots_list(...)))
+  newAttribs <- dropNullsOrEmpty(dots_list(...))
+  if (any(names2(newAttribs) == "")) {
+    warning("At least one of the new attribute values did not have a name. Did you forget to include an attribute name?")
```
cpsievert, line 297: I feel like we should `stop()` here since the error you'll eventually get isn't useful

R/tags.R
```diff
@@ -492,6 +496,8 @@ tagWrite <- function(tag, textWriter, indent=0, eol = "\n") {
 
   # write attributes
   for (attrib in names(attribs)) {
+    # Can not display attrib without a key
+    if (identical(attrib, "")) next
```
cpsievert, line 500: Oh, I see, so this would 'fix' this situation?

```r
x <- div()
x$attribs[[1]] <- "a"
```

cpsievert: Part of me thinks that we should also `stop()` here with an informative error

schloerke: Correct. No error is thrown.

Tested below

R/tags.R
```diff
@@ -492,6 +499,13 @@ tagWrite <- function(tag, textWriter, indent=0, eol = "\n") {
 
   # write attributes
   for (attrib in names(attribs)) {
+    # Can not display attrib without a key
+    if (identical(attrib, "")) {
```
cpsievert, line 503: Pretty sure this would be faster? Or do we have to be considerate of 

```suggestion
    if (!nzchar(attrib)) {
```

schloerke: Using a combo of `names2(attribs)` with `attrib == ""`

R/tags.R
```diff
@@ -292,7 +292,14 @@ tagFunction <- function(func) {
 #' @export
 tagAppendAttributes <- function(tag, ...) {
   throw_if_tag_function(tag)
-  tag$attribs <- c(tag$attribs, dropNullsOrEmpty(dots_list(...)))
+  newAttribs <- dropNullsOrEmpty(dots_list(...))
+  if (any(names2(newAttribs) == "")) {
```
cpsievert, line 296: Just to be consistent

```suggestion
  if (any(!nzchar(names2(newAttribs)))) {
```

cpsievert (approved):

NEWS.md
```diff
@@ -8,6 +8,8 @@
 
 * Closed #197: Fixed rendering of boolean attributes in <script> tags rendered via renderDependencies() (#197, thanks @atusy).
 
+* Closed #222; Unnamed attributes are no longer allowed to be appended via `tagAppendAttribs()`. When trying to print unnamed tag attribs, a better error message is provided. (#229)
```
cpsievert, line 11:
```suggestion
* Closed #222: Unnamed attributes are no longer allowed to be appended via `tagAppendAttribs()`. When trying to print unnamed tag attribs, a better error message is provided. (#229)
```

PR #230, opened 2021-04-21
`tagQuery()`: Rename`$root()` to `$allTags()`, `$selected()` to `$selectedTags()`; Print `$selectedTags()` like a `list()`

schloerke:
- [x] @wch, should we adjust the print method of `tagList()` to display `[[i]]` like a regular list?
  * If so, we can remove `visibleTagList()` 

```r
tagQuery(div(span(), a()))$root()
#> <div>
#>   <span></span>
#>   <a></a>
#> </div>

tagQuery(div(span(), a()))$selected()
#> [[1]]
#> <div>
#>   <span></span>
#>   <a></a>
#> </div>

# has extra `visible` class
tagQuery(div(span(), a()))$selected() %>% str()
#> List of 1
#>  $ :List of 3
#>   ..$ attribs : Named list()
#>   ..$ name    : chr "div"
#>   ..$ children:List of 2
#>   .. ..$ :List of 3
#>   .. .. ..$ attribs : Named list()
#>   .. .. ..$ name    : chr "span"
#>   .. .. ..$ children: list()
#>   .. .. ..- attr(*, "class")= chr "shiny.tag"
#>   .. ..$ :List of 3
#>   .. .. ..$ attribs : Named list()
#>   .. .. ..$ name    : chr "a"
#>   .. .. ..$ children: list()
#>   .. .. ..- attr(*, "class")= chr "shiny.tag"
#>   ..- attr(*, "class")= chr "shiny.tag"
#>  - attr(*, "class")= chr [1:3] "htmltools.tag.list.visible" "shiny.tag.list" "list"

tagQuery(div(span(), a()))$root() %>% format() %>% cat()
#> <div>
#>   <span></span>
#>   <a></a>
#> </div>
```


R/tag_query.R
```diff
@@ -985,17 +997,11 @@ wrapWithRootTag <- function(x) {
...
-  len <- length(children)
-  if (len == 1) {
-    children[[1]]
-  } else if (len > 1) {
-    do.call(tagList, children)
-  } else {
-    # no children?
-    NULL
-  }
+  do.call(visibleTagList, children)
```
cpsievert, line 1000: Could this just be `tagList(root$children)`? I'm not in love with the fact that you get the `[[` indexing for `Root`.

cpsievert: Also a comment about `root` being a pseudo-tag might be helpful here. Also, perhaps it makes sense to drop the pseudo-tag data structure for a tagList() ?

schloerke: > perhaps it makes sense to drop the pseudo-tag data structure for a tagList()?

We can't easily edit a tagList in place. We can easily edit a tag env in place. A pseudo root is preferred over `<<-`.  

-----------------

Using a pseudo root tag helps with trying to make a new tagQuery object from a set of tagEnvs. For each tagEnv, we can look for the pseudoRoot to know they come from the same original tagQuery object. If all tagEnvs have the same pseudoRoot tag env, then we can create a new tagQuery object with the selected elements.  If there is a mix of root tags, then an error is thrown.  

This check would not be possible without a pseudoRoot tag env.

wch:
Why do we want to always return a `tagList`? I do see that there's an advantage in that the return type is more stable, but on the other hand, in most cases, the input is a single tag object at the top level, and the user would expect the returned object to also be a tag (instead of `tagList`).

schloerke:
Thought process was "anything in gives a tagList out. Then the same tagList in gives a tagList out." After one cycle, the input and output objects are stable.

I'm ok putting it back to the flexible return type knowing that it is not stable as I would believe that few people will call `tagList(myTag)$append(otherTag)$root()` and not be shocked by the return type changing.

wch:
I think that it's OK to return a variable type. If anything, I think always returning a `tagList` will cause more confusion, because in most cases, that would mean that what comes out is different from what goes in. If we really wanted, we could add a parameter to `$root()` which tells it to always return a `tagList`, and have it default to `FALSE`.

Also, as I was looking at this PR and https://github.com/rstudio/shiny/pull/3372, I think that name `$root()` is a bit confusing. It sort of suggests you're just getting the root node, but you're really getting the whole tree. Not sure of a better alternative at this moment, though.

schloerke:
@wch 

> I think that name $root() is a bit confusing

* `$dom()`?
* `$topLevelTags()`?
* `$asTags()`?

`$html()` is too [related to `.html()`](https://api.jquery.com/html/#html) which has the description of ...

> Get the HTML contents of the first element in the set of matched elements.

... so my gut reaction is not comfortable using `$html()` due to the differences in implementation.

R/tag_query.R
```diff
@@ -339,6 +339,21 @@ assertNotTagEnvLike <- function(x, fnName) {
 }
 
 
+# Shim in a class so that the print method of tagList() is not used
+# Yet knit print methods will work as if they are tagList objects.
+visibleTagList <- function(...) {
+  y <- tagList(...)
+  oldClass(y) <- c("htmltools.tag.list.visible", oldClass(y))
```
wch, line 346: I think that this should be `shiny.tag.list.visible`, to be consistent with the other class names we use. If we could go back in time and change them all to `htmltools.tag.*`, that would be nice, but we can't.

wch: Also, `"visible"` doesn't feel like quite the right name, and I feel that it's a little weird to add a new class just so that it'll print differently. How about just adding an attribute to the tagList, like `print_list=TRUE`, and then checking for that attribute in `print.shiny.tag.list`? (Note that `print_list` isn't a great name.)

wch:
Some possible names:

* `allTags()`
* `asTags()`. This might be a little confusing because users may expect it to just return the selected items.  Maybe `$selected()` should also then be renamed to `$asTagsSelected()` or `$selectedAsTags()`? That one doesn't need a short, concise name because it will not be commonly used, I think.

schloerke:
How about `$allTags()` and `$selectedTags()`?

R/tag_query.R
schloerke: Rename: ✅ 

"Visible": I'm going to have `$selected()` (which is where `visibleTagList()` was called) return a `list()` and not a `tagList()`. This will solve our complications. (`$root()` will still return a `tagList` and not a `list()`)

schloerke: 😞 I now remember why `visibleTagList()` was used. Going to use a `print.as.list = TRUE` attribute

wch: Just curious: why doesn't it work to return a `list()`?

wch:
`$allTags()` and `$selectedTags()` works for me. @cpsievert?

We should also have a version bump because of the new names.

R/tag_query.R
schloerke: Rename: bd66aa4
`print.as.list` attr: 4f6cc0e

cpsievert:
LGTM 👍

R/tags.R
```diff
@@ -211,7 +211,15 @@ as.character.html <- function(x, ...) {
 }
 
 #' @export
-print.shiny.tag.list <- print.shiny.tag
+print.shiny.tag.list <- function(x, ...) {
+  if (isTRUE(attr(x, "print.as.list", exact = TRUE))) {
+    attr(x, "print.as.list") <- NULL
+    class(x) <- setdiff(class(x), "shiny.tag.list")
+    return(print(x))
```
wch, line 218: Does it work if you just call `return(NextMethod())` instead of these lines?

wch (approved):

cpsievert (approved):

R/tags.R
schloerke: It does work, but it has ugly print output...

```r
tagQuery(div())$selectedTags()
#> [[1]]
#> <div></div>
#> 
#> attr(,"class")
#> [1] "shiny.tag.list" "list"
#> attr(,"print.as.list")
#> [1] TRUE
```

-------------------

Going to keep it as is

Issue #216, opened 2021-04-19
tagQuery()'s $addClass() should allow NULL values


cpsievert:
Which would be a no-op and is convenient for this sort of use:

```r
tagQuery(div())$addClass(if (F) "a")
#> Error in getCssClass(class) : 
#>  `class` must resolve to a character value with a length of at least 1
```

```r
tagAppendChildren(div(), class = if (F) "a")
#> <div></div>
```

PR #224, opened 2021-04-20
`tagQuery()` docs; Add `.cssSelector` parameter to tag methods

cpsievert:
Closes https://github.com/rstudio/htmltools/issues/218
Closes https://github.com/rstudio/htmltools/issues/234
Closes https://github.com/rstudio/htmltools/issues/251

PR #236, opened 2021-04-22
Allow tag query `$*Class()` methods to no-op on length 0 inputs

schloerke:
Fixes https://github.com/rstudio/htmltools/issues/216

Also works with `character(0)` inputs (or any length 0 inputs)

--------------

Moved the `$hasClass()` method down so that the method order is "add, empty-like, has", similar to the attr method on `tagQuery()`

### DIFF
--- a/R/tag_query.R
+++ b/R/tag_query.R
@@ -732,18 +732,18 @@ tagQuery_ <- function(
           tagQueryClassRemove(selected_, class)
           invisible(self)
         },
-        #' * `$hasClass(class)`: Determine whether the selected elements have a
-        #' given class. Returns a vector of logical values.
-        hasClass = function(class) {
-          rebuild_()
-          tagQueryClassHas(selected_, class)
-        },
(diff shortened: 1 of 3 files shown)


R/tag_query.R
```diff
@@ -732,19 +732,19 @@ tagQuery_ <- function(
...
-          tagQueryClassHas(selected_, class)
-        },
         #' * `$toggleClass(class)`: If the a class value is missing, add it. If
         #' a  class value already exists, remove it.
         toggleClass = function(class) {
           rebuild_()
           tagQueryClassToggle(selected_, class)
           invisible(self)
         },
+        #' * `$hasClass(class)`: Determine whether the selected elements have a
```
cpsievert, line 742: Would you mind moving this back so it doesn't clash with #224?

cpsievert (approved):

R/tag_query.R
schloerke: 👍 

Mind adding this in #224?

PR #265, opened 2021-05-20
`tagQuery()`: Do not adjust the class attrib position when adding a class

schloerke:
Previous approach: Remove all `class` values, Add the new `class` value.

New approach: 
* If no `class` values
	* Add the new `class` value
* Else
	* If more than one `class` values
		* Remove the extra occurrences of `class`
	* Overwrite the first `class` value

With the new approach, the first class position is kept, preserving the attribute order

### DIFF
--- a/tests/testthat/test-tag-query.R
+++ b/tests/testthat/test-tag-query.R
@@ -835,7 +835,7 @@ test_that("adding a class does not reorder attribs", {
     div(class="bar foo", test = "A", "text")
   )
 
-  # Multiple classes class
+  # Multiple classes
   expect_equal_tags(
     tagQuery(div(class = "bar", test = "A", class = "baz", "text"))$addClass("foo")$allTags(),
     div(class = "bar baz foo", test = "A", "text")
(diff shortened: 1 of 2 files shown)


tests/testthat/test-tag-query.R
```diff
@@ -821,7 +821,26 @@ test_that("tag methods do not unexpectedly alter tag envs", {
...
+  expect_equal_tags(
+    tagQuery(div(class = "bar", test = "A", "text"))$addClass("foo")$allTags(),
+    div(class="bar foo", test = "A", "text")
+  )
+
+  # Multiple classes
+  expect_equal_tags(
+    tagQuery(div(class = "bar", test = "A", class = "baz", "text"))$addClass("foo")$allTags(),
+    div(class = "bar baz foo", test = "A", "text")
+  )
```
cpsievert, line 842: `expect_equal_tags()` will fail if the attribute ordering is different, right?

schloerke: Correct

cpsievert (approved):

PR #267, opened 2021-05-21
Implement post render hooks; add hooks to tagList()

cpsievert:
(I'm not 100% sure this is an avenue we want to pursue, but it does seem like a decent way to solve `shiny::bootstrapLib()'`s static rendering issues)

This main point of this PR is to add a `.postRenderHook` -- the primary motivation for it is to provide a way to clean-up any side-effects that you may want to perform with `.renderHook`. Also, since `bootstrapLib()` is an `htmlDependency()` (not a tag), I've also added the ability to attach hooks to `tagList()` (this will allow us to fully deprecate `tagFunction()` since we can now stop using it in `shiny::bootstrapLib()`).

Technically this is a breaking change since we'll now disallow render hooks to return objects that require `as.tags()` to be written to HTML, but that seems like a fair requirement to me.


R/tags.R
```diff
@@ -382,22 +395,24 @@ tagFunction <- function(func) {
 #'   tags$p("Something else")
 #' })
 #' newObj
-tagAddRenderHook <- function(tag, func, replace = FALSE) {
+tagAddRenderHook <- function(tag, func, replace = FALSE, post = FALSE) {
```
cpsievert, line 398: Add `tagAddPostRenderHook` instead

R/tags.R
```diff
@@ -323,10 +333,13 @@ tagFunction <- function(func) {
...
-#' @param tag A [`tag()`] object.
-#' @param func A function (_hook_) to call when the `tag` is rendered. This function
-#'   should have at least one argument (the `tag`) and return anything that can
-#'   be converted into tags via [as.tags()].
+#' @param tag A [tag()] or [tagList()].
+#' @param func A function (_hook_) to call when the `tag` is rendered. This
+#'   function should have at least one argument (the `tag`) and it's return
+#'   value should be 'resolved' in the sense that it shouldn't return:
+#'   * A [tagFunction()].
+#'   * A [tag()] or [tagList()] with render hooks.
```
cpsievert, line 341: It'd be nice to support this (by doing `tagify(as.tags(hook(ui)))`?)

R/tags.R
```diff
@@ -1182,12 +1197,31 @@ withTags <- function(code) {
...
+
+tagify_node <- function(ui) {
+  if (isTag(ui) || isTagList(ui)) {
+    pre <- attr(ui, "renderHooks")
+    post <- attr(ui, "postRenderHooks")
+    attr(ui, "renderHooks") <- NULL
+    attr(ui, "postRenderHooks") <- NULL
+
+    for (hook in pre) {
+      ui <- hook(ui)
```
wch, line 1211: Wrap this in a `tryCatch()` and warn on errors.

R/tags.R
```diff
@@ -1182,12 +1197,31 @@ withTags <- function(code) {
...
+    }
+    for (hook in post) {
+      on.exit(return(hook(ui)), add = TRUE)
+    }
```
wch, line 1215:
```suggestion
    on.exit({
      for (hook in post) {
        tryCatch(
	      ui <- hook(ui),
	      error = function(e) { warning(".....") }
	    )  
      }
      
      # Does this return work as expected?
      return(ui)
    }
```

cpsievert:
@wch I’ve got something that’s at least somewhat working now, but unfortunately I don’t see a way for `.postRenderHook` to modify its input -- do you think that'll be an issue and/or see a way to support that?

R/tags.R
```diff
@@ -1182,14 +1204,37 @@ withTags <- function(code) {
...
+
+    if (isTag(ui) || isTagList(ui)) ui else tagify(as.tags(ui))
+
   }, FALSE)
 }
 
+tryWarn <- function(expr) {
+  tryCatch(expr, error = function(e) warning(conditionMessage(e)))
+}
```
wch, line 1236: I think we'll want a better warning message here. For example:

```
> tryWarn(x <- stop("xyz"))
Warning message:
In value[[3L]](cond) : xyz
```


Maybe something like this:

```R
tryWarn <- function(expr) {
  result <- NULL
  error_message <- NULL
  result <- tryCatch({ expr }, error = function(e) error_message <<- conditionMessage(e))
  
  if (is.null(error_message)) {
    return(result)
  } else {
    warning(error_message)
    invisible(NULL)
  }
}

tryWarn(x <- stop("xyz"))
#> Warning message:
#> In tryWarn(x <- stop("xyz")) : xyz
```

However, one kind of strange thing about this is that in the case of an error, it (invisibly) returns `NULL`. I'm not sure what a better return value would be, though.

R/tags.R
```diff
@@ -1182,14 +1204,37 @@ withTags <- function(code) {
...
+    attr(ui, "renderHooks") <- NULL
+
+    post <- attr(ui, "postRenderHooks")
+    if (length(post)) {
+      postHooks <<- c(postHooks, post)
+      attr(ui, "postRenderHooks") <- NULL
+    }
+
+    if (isTag(ui) || isTagList(ui)) ui else tagify(as.tags(ui))
+
```
wch, line 1230: A couple things feel strange about this. As mentioned elsewhere, this doesn't use the return value from the post hook, but I think that should be possible.

One thing that feels strange is the asymmetry between how the pre and post hooks are run. The pre hooks are run from inside the anonymous function passed to `rewriteTags()`; in contrast, the post hooks are *collected* within the anonymous function, but they're *executed* by `tagify`'s `on.exit()`. Could the call to `on.exit()` be moved into the inner anonymous function?

Or could the invocation of both the pre and post render hooks happen outside of the anonymous function?

wch:
We'll need some good unit tests to make sure the pre and post hooks are executed properly.

R/tags.R
cpsievert: > Could the call to on.exit() be moved into the inner anonymous function?

I don't think that's quite right since in that case the post hooks could run too early in some cases.

tests/testthat/test-tag-hooks.R
```diff
@@ -0,0 +1,224 @@
...
+          message("C", state)
+        }
+      ),
+      .postRenderHook = function() {
+        state <<- state + 1
+        message("D", state)
+      }
+    )
+  ))
+  expect_equal(state, 4)
```
cpsievert, line 169:
```suggestion
  state <- 0
  renderTags(tagList(
    div(
      .renderHook = function(x) {
        state <<- state + 1
        message("A", state)
      },
      div(
        .renderHook = function(x) {
          state <<- state + 1
          message("B", state)
        },
        .postRenderHook = function() {
          state <<- state + 1
          message("C", state)
        }
      ),
      .postRenderHook = function() {
        state <<- state + 1
        message("D", state)
      }
    )
  ))
  expect_equal(state, 4)
```

tests/testthat/test-tag-hooks.R
```diff
@@ -0,0 +1,224 @@
...
+
+
```
cpsievert, line 171:
```suggestion
```

PR #278, opened 2021-07-08
Relax the required version of rlang

cpsievert:
`rlang::obj_address()` won't be released until at least August


R/tag_query.R
```diff
@@ -1,7 +1,11 @@
 #' @importFrom fastmap fastmap faststack
-#' @importFrom rlang obj_address
 NULL
 
+# TODO: change to rlang::obj_address() after rlang > 0.4.11 hits CRAN
```
schloerke, line 4:
```suggestion
# TODO: change to rlang::obj_address() after rlang >= 0.4.12 hits CRAN
```

schloerke (approved):

wch:
Have you tested the performance of this? IIRC, the reason we used `obj_address()` was to be fast, but in this new version, `getFromNamespace()` is likely to be slow.

cpsievert:
I don't have a good benchmark in mind, but even if it is slow, would it be worth delaying release?

schloerke:
We can set the local variable to NULL. Then in .onLoad, we can call get from namespace once and set the value with a <<-  

This allows for rlang to update after htmltools is installed

wch:
I wouldn't delay the release for it, but there are other ways of getting the function that are faster. For example, it could use `.getNamespace()`. Or it could fetch `sexp_address` the first time `obj_address` is called, and cache it. That's actually probably the best way to go. 

```R
obj_address_cached <- local({
  sexp_address <- NULL
  
  function(x) {
    if (is.null(sexp_address)) {
      sexp_address <<- getFromNamespace("sexp_address", "rlang")
    }
    sexp_address(x)
  }  
})
  
x <- 1:10
  
microbenchmark::microbenchmark(
  getFromNamespace("sexp_address", "rlang")(x),
  .getNamespace("rlang")$sexp_address(x),
  obj_address_cached(x)
)
#> Unit: microseconds
#>                                          expr   min     lq    mean median     uq    max neval
#>  getFromNamespace("sexp_address", "rlang")(x) 4.343 4.4815 4.87278 4.5785 4.7365 24.332   100
#>        .getNamespace("rlang")$sexp_address(x) 1.259 1.3100 1.43425 1.3875 1.4650  3.989   100
#>                         obj_address_cached(x) 1.201 1.2575 1.43651 1.3225 1.4070 11.206   100
```

R/tags.R
```diff
@@ -54,6 +54,19 @@ registerMethods <- function(methods) {
...
   ))
+
+  # TODO: After rlang > 0.4.11 hits CRAN, remove this and replace
+  # with ` #' @importFrom rlang obj_address`
+  # (lionel says rlang:::sexp_address() will be available for the next few years)
+  assign_obj_address()
+  setHook(
+    packageEvent("rlang", "onLoad"),
+    function(...) assign_obj_address()
+  )
```
wch, line 65: This isn't be necessary, and would actually throw an error if it were to execute. It's not necessary because we can assume that rlang is available when `.onLoad` is called, since it's listed in Imports. As for the error: since the namespace is locked after the `.onLoad` callback executes, the hook's `assign()` would fail after that point.

R/tags.R
```diff
@@ -54,6 +54,19 @@ registerMethods <- function(methods) {
...
+}
+
+assign_obj_address <- function() {
+  assign("obj_address", getFromNamespace("sexp_address", "rlang"), environment(.onLoad))
 }
```
wch, line 70: After removing the `setHook()`, may as well move this into `.onLoad`, since it only needs to be  called once.

PR #289, opened 2021-08-24
Update workflows to use latest config and dynamic R versions

schloerke:


schloerke:
Closing in favor of #296

PR #298, opened 2021-11-08
Use faster versions of system.file() provided by staticimports

cpsievert:


cpsievert:
@wch the CI should fix itself once https://github.com/wch/staticimports/pull/11 is merged

PR #302, opened 2021-12-06
`tagQuery()`: Relocate html deps to child objects

### LINKED ISSUE #301: `tagQuery()`: Nested `tagList()` `htmlDependency()` values are being dropped
Hi, I would like to use  `tagQuery()` to create/modify tags like cards, navbars and panels that serve as containers for UI output.
I've already posted a detailed explanation of my problem including a reprex here:
https://community.rstudio.com/t/shiny-htmltools-custom-tags-behave-weird-when-using-tagquery-to-modify-tags/123308

Long story short, running  `tagQuery()` on a tag, that somewhere contains an htmlwidget UI output, changes the structure of the tag, such that the UI output won't be shown anymore.
Weirdly, this is only the case, when you do this at least twice.

I hope you can tell what the problem is and help me out.

Thanks in advance and kind regards
David

schloerke:
Fixes #301 
cc @DavidJesse21

- [x] test added
- [x] news entry
- [x] document that tag query will relocate all htmldeps after usage (and also flatten nested lists in children)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,7 +2,7 @@
 
 ## Bug fixes
 
-* Closed #301: `tagQuery()` was failing to copy all `tagList()` html dependencies within nest child tag lists. (#302)
+* Closed #301: `tagQuery()` was failing to copy all `tagList()` html dependencies within nest child tag lists. `tagQuery()` will now relocate html dependencies as child objects. (#302)
 
 * Closed #290: htmltools previously did not specify which version of fastmap to use, and would fail to install with an old version of fastmap. (#291)
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,8 @@
 
 ## Bug fixes
 
+* Closed #301: `tagQuery()` was failing to copy all `tagList()` html dependencies within nest child tag lists. (#302)
+
 * Closed #290: htmltools previously did not specify which version of fastmap to use, and would fail to install with an old version of fastmap. (#291)
 
 * `copyDependencyToDir()` no longer creates empty directories for dependencies that do not have any files. (@gadenbuie, #276)
--- a/R/tags.R
+++ b/R/tags.R
@@ -1292,6 +1292,14 @@ flattenTagsRaw <- function(x) {
       ret <- unlist(lapply(x, flattenTagsRaw), recursive = FALSE)
       # Copy over attributes put on the original list (ex: html deps)
       mostattributes(ret) <- attributes(x)
+
+      # Copy over individual html deps into the final list
+      # This does not work for attributes in general,
+      # but we can make it work for html deps
+      x_html_deps <- unlist(lapply(x, htmlDependencies), recursive = FALSE)
+      if (length(x_html_deps) > 0) {
+        ret <- attachDependencies(ret, x_html_deps, append = TRUE)
+      }
       ret
     }
   } else {
--- a/R/tags.R
+++ b/R/tags.R
@@ -1280,29 +1280,38 @@ flattenTags <- function(x) {
 # By not calling `as.tags(x)`, tagFunctions are not evaluated and other items
 # are not converted.
 flattenTagsRaw <- function(x) {
-  if (isTag(x) || isTagEnv(x)) {
+  if (isTagEnv(x)) {
+    # For tags, wrap them into a list (which will be unwrapped by caller)
+    list(x)
+  } else if (isTag(x)) {
+    tagDeps <- htmlDependencies(x)
+    # Move the tag's html deps to its children
+    if (length(tagDeps) > 0) {
+      x$children <- c(x$children, tagDeps)
+      htmlDependencies(x) <- NULL
+    }
     # For tags, wrap them into a list (which will be unwrapped by caller)
     list(x)
   } else if (isTagList(x)) {
-    if (length(x) == 0) {
-      # Empty lists are simply returned
-      x
-    } else {
-      # For items that are lists (but not tags), recurse
-      flattened_tags <- lapply(x, flattenTagsRaw)
-      ret <- unlist(flattened_tags, recursive = FALSE)
-      # Copy over attributes put on the original list (ex: html deps)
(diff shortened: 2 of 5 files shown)


wch:
I'm not saying we need to do this now, but there's another possible way to approach this which may simplify things in the long run: in the tag to tag-environment conversion, convert all html dependency attributes into children. This would provide a normalized data structure where the html deps are always children, instead of sometimes being children and sometimes being attributes.

This does mean that when the tag env object is converted back into a tag, the structure will be changed, but I think that already happens in other ways.

If we want to push this even further, it may make sense to always make html deps children, even on regular tag objects. For example, `attachDependencies()` would simply add the dependencies as children, instead of as attributes. I haven't thought about this enough yet, though, to say whether or not there would be unexpected breakage from doing this. If we could always use a normalized format, it would simplify our code in the long run; the only issue is if whether it would cause problems in the short term.

cpsievert:
> If we want to push this even further, it may make sense to always make html deps children, even on regular tag objects. For example, attachDependencies() would simply add the dependencies as children, instead of as attributes.

FWIW, I vaguely recall going down this road for something else and I'm pretty sure I backed out because it was going to break a lot of stuff (maybe because lots of people were using `attr(x, "html_dependencies")` directly?)

schloerke:
While I like the list item structure as well, I agree with @cpsievert . Moving to a list item completely breaks retrieving html deps using `htmlDependencies(x)`

-------------------

> If we could always use a normalized format, it would simplify our code in the long run; the only issue is if whether it would cause problems in the short term.

I'm totally up for standardization, but it will cause breakages in the near term.

wch:
I think it's arguable that the current behavior of `htmlDependencies()` is incorrect, when it comes to dealing with deps that are added as children:

```R
library(htmltools)
x <- div("hello", htmlDependency("foo", "1.0", "/foo"))
htmlDependencies(x)
#> NULL
```

`findDependencies()` knows how to handle them, though:

```R
findDependencies(x)
#> [[1]]
#> List of 10
#>  $ name      : chr "foo"
#>  $ version   : chr "1.0"
#>  $ src       :List of 1
#>   ..$ file: chr "/foo"
#>  $ meta      : NULL
#>  $ script    : NULL
#>  $ stylesheet: NULL
#>  $ head      : NULL
#>  $ attachment: NULL
#>  $ package   : NULL
#>  $ all_files : logi TRUE
#>  - attr(*, "class")= chr "html_dependency"
```

In this search I see some cases where the code accesses `attr(x, "html_dependencies")`, but not a prohibitively amount:
  https://github.com/search?q=org%3Acran+html_dependencies&type=code

At any rate, this isn't something we necessarily need to do in this PR, but if not, we should open another issue about it.

wch (approved):
We can merge this, but should also file an issue referring back to the discussion here, about normalizing the html dependency data structures in general.

DavidJesse21:
Hey, first of all thank you all for catching up on this issue!
I have now downloaded the dev version of `{htmltools}` via:
```r
remotes::install_github("rstudio/htmltools", ref = "taglist_squash")
```

Without looking into details I cannot notice any differences regarding my example given [here](https://community.rstudio.com/t/shiny-htmltools-custom-tags-behave-weird-when-using-tagquery-to-modify-tags/123308).
Have I just done something wrong downloading the latest version with the fixes or how can you explain that my problem preserves?

Cheers
David

schloerke:
Dang. Still missed some. 😞 Will explore more soon. 
@DavidJesse21 Your code is great. You are doing nothing wrong.

------------------

Maybe we use @wch's approach on the HTML deps for anything that goes through `tagQuery()`. Keeping track of all the dep attributes attributes when subsetting (`[[`), `unlist()`ing, and flattening children is becoming overly difficult.

schloerke:
@DavidJesse21 Can you check again? I have confirmed it on my side, but want to be sure.

------------------

Changes: 
* When sending a set of tags to `tagQuery()`:
	* All tag html deps (in attrs) will be appended to the tag's `$children` field. 
	* All tag lists will remove all html dependencies (in attrs) and append them to the end of the list

This works recursively and will expand on the _children_ slots and remove the brittleness of attributes when flattening all `$children` fields.

This change seems warranted as the `$children` structure was not guaranteed to be the same after sending to `tagQuery()`. Now, this also extends to the location of the html dependencies.  

The agreement of `tagQuery()` still stands that when **rendering** the html tags both the input tags and output tags (of `tagQuery()`) should render the same result. (The internal structures will probably change.)

DavidJesse21:
This seems to work now, thank you :)

R/tags.R
```diff
@@ -1280,20 +1280,38 @@ flattenTags <- function(x) {
...
+    # Instead, use the _newer_/stable approach of adding the html dep as a direct child
+    tagListDeps <- htmlDependencies(ret)
+    # Append the incoming tagList's html deps to its children
+    tagListDepsLen <- length(tagListDeps)
+    if (tagListDepsLen > 0) {
+      # Do position insert to not lose attrs on ret
+      retLen <- length(ret)
+      ret[(retLen + 1):(retLen + tagListDepsLen)] <- tagListDeps
+      # Remove html deps on ret, as they are now in the children
+      htmlDependencies(ret) <- NULL
```
cpsievert, line 1312: Maybe wrap this in a helper function and use it above on new line 1287 (this is equivalent, but faster, right)?

schloerke: Great catch! Would prolly have lost html deps on the tag children if the old code was kept.

wch (approved):

NEWS.md
```diff
@@ -2,6 +2,8 @@
 
 ## Bug fixes
 
+* Closed #301: `tagQuery()` was failing to copy all `tagList()` html dependencies within nest child tag lists. (#302)
```
schloerke, line 5:
```suggestion
* Closed #301: `tagQuery()` was failing to copy all `tagList()` html dependencies within nest child tag lists. `tagQuery()` will now relocate html dependencies as child objects. (#302)
```

PR #309, opened 2021-12-29
fix: copyDependencyToDir tries to copy non-existent files unintentionally

atusy:
when script is a list and `all_files` is `FALSE`

This happens because `copyDependencyToDir` considers `unlist(dependency[c('script', 'stylesheet', 'attachment')])` as paths to files, but script can be a list with some html attributes such as `defer`.

``` r
library(htmltools)

# Prepare example data
src <- tempfile()
dir.create(src, showWarnings = FALSE)
script <- c("1.js", "2.js")
invisible(file.create(file.path(src, script)))
deps <- htmlDependency(
  name = "example",
  version = "1.0",
  src = src,
  #script = script,
  script = lapply(script, function(x) list(src = x, defer = NA)),
  all_files = FALSE
)

# Try copyDependencyToDir
outputDir = tempfile()
dir.create(outputDir, showWarnings = FALSE)
copyDependencyToDir(deps,outputDir = outputDir)
#> Error in htmltools::copyDependencyToDir(deps, outputDir = outputDir): Can't copy dependency files that don't exist: '/tmp/RtmpzXKs5M/file121c6ec550eb/1.js', '/tmp/RtmpzXKs5M/file121c6ec550eb/NA', '/tmp/RtmpzXKs5M/file121c6ec550eb/2.js', '/tmp/RtmpzXKs5M/file121c6ec550eb/NA'
```

<sup>Created on 2021-12-29 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

### DIFF
--- a/R/html_dependency.R
+++ b/R/html_dependency.R
@@ -355,7 +355,10 @@ copyDependencyToDir <- function(dependency, outputDir, mustWork = TRUE) {
   dependency$src$file <- normalizePath(target_dir, "/", TRUE)
 
   files <- if (dependency$all_files) list.files(dir) else {
-    unlist(dependency[c('script', 'stylesheet', 'attachment')])
+    c(
+      src_script(dependency[["script"]]),
+      unlist(dependency[c('stylesheet', 'attachment')])
+    )
   }
 
   if (length(files) == 0) {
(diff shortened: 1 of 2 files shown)


cpsievert:
Addressed by https://github.com/rstudio/htmltools/commit/5fe6cd67b25153302351fd026a070d1a7660d7a6

PR #321, opened 2022-03-31
Don't assume all scripts are given as character variables.  

### LINKED ISSUE #320: htmlDependency() sometimes doesn't allow list entries for `script`
The `script` component of `htmlDependency` can be a list containing list entries in order to set attributes on the included script.  However, if `all_files = FALSE` is used, this fails.  For example, the markdown file below fails to render.  Changing `all_files` to `TRUE` makes it work, as does changing `self_contained` to `true`.

The problem appears to be here:

https://github.com/rstudio/htmltools/blob/e47232123e8916bf7c8d95f4513d0fa77c8f4d13/R/html_dependency.R#L358

because the code assumes dependencies are all simple filenames.

Here's the sample file to demonstrate:

``````
---
title: "Untitled"
output: 
  html_document:
    self_contained: false
---

```{r}
javascript <- 'console.log("this is Javascript to write to the console");'
f1 <- tempfile(fileext = ".js")
writeLines(javascript, f1)

text <- 'this is some text to go in a script'
f2 <- tempfile(fileext = ".txt")
writeLines(text, f2)
```

```{r}
library(htmltools)
deps <- htmlDependency("test", "1.0",
                       src = tempdir(),
                       script = list(basename(f1),
                                     list(src = basename(f2),
                                          type = "text/plain",
                                          id = "shader")),
                       all_files = FALSE
                       )
tag <- tags$p("this is visible text", deps)
tag
```
``````

### PR DESCRIPTION
**Pull request by dmurdoch**
Also fixes the error message to only report missing files.  Fixes #320.

### DIFF
--- a/R/html_dependency.R
+++ b/R/html_dependency.R
@@ -356,17 +356,10 @@ copyDependencyToDir <- function(dependency, outputDir, mustWork = TRUE) {
 
   if (dependency$all_files)
     files <- list.files(dir)
-  else {
-    file_list <- c(dependency[['script']],
-                   dependency[['stylesheet']],
-                   dependency[['attachment']])
-    if (anyUnnamed(file_list))
-      files <- vapply(file_list,
-                      function(f) if (is.list(f)) f[['src']] else f,
-                      "")
-    else
-      files <- file_list[['src']]
-  }
+  else
+    files <- c(pluck_src(dependency$script),
+               pluck_src(dependency$stylesheet),
+               pluck_src(dependency$attachment))
 
   if (length(files) == 0) {
     # This dependency doesn't include any files
--- a/R/html_dependency.R
+++ b/R/html_dependency.R
@@ -354,9 +354,11 @@ copyDependencyToDir <- function(dependency, outputDir, mustWork = TRUE) {
   dir.create(target_dir)
   dependency$src$file <- normalizePath(target_dir, "/", TRUE)
 
-  files <- if (dependency$all_files) list.files(dir) else {
-    unlist(dependency[c('script', 'stylesheet', 'attachment')])
-  }
+  files <- if (dependency$all_files)
+    list.files(dir)
+  else
+    vapply(unlist(dependency[c('script', 'stylesheet', 'attachment')], recursive = FALSE),
+           function(f) if (is.list(f)) f[['src']] else f, "")
 
   if (length(files) == 0) {
     # This dependency doesn't include any files
@@ -366,11 +368,11 @@ copyDependencyToDir <- function(dependency, outputDir, mustWork = TRUE) {
   }
 
   srcfiles <- file.path(dir, files)
-  if (any(!file.exists(srcfiles))) {
+  if (any(bad <- !file.exists(srcfiles))) {
     stop(
       sprintf(
         "Can't copy dependency files that don't exist: '%s'",
-        paste(srcfiles, collapse = "', '")
+        paste(srcfiles[bad], collapse = "', '")
       )
     )
   }
--- a/R/html_dependency.R
+++ b/R/html_dependency.R
@@ -369,11 +369,12 @@ copyDependencyToDir <- function(dependency, outputDir, mustWork = TRUE) {
   }
 
   srcfiles <- file.path(dir, files)
-  if (any(bad <- !file.exists(srcfiles))) {
+  missing_srcfiles <- !file.exists(srcfiles)
+  if (any(missing_srcfiles)) {
     stop(
       sprintf(
         "Can't copy dependency files that don't exist: '%s'",
-        paste(srcfiles[bad], collapse = "', '")
+        paste(srcfiles[missing_srcfiles], collapse = "', '")
       )
     )
   }
--- a/R/html_dependency.R
+++ b/R/html_dependency.R
@@ -354,11 +354,19 @@ copyDependencyToDir <- function(dependency, outputDir, mustWork = TRUE) {
   dir.create(target_dir)
   dependency$src$file <- normalizePath(target_dir, "/", TRUE)
 
-  files <- if (dependency$all_files)
-    list.files(dir)
-  else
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by cpsievert** [comment]
Thanks @dmurdoch! If you don't mind, it'd also be great to have a unit test

**Comment by dmurdoch** [comment]
Sure, I'll see if I can add one.

**Comment by dmurdoch** [comment]
The last commit adds a test.  It just tries to see if the function runs without error; I think other tests check that the results are okay.

File: R/html_dependency.R [hunk 1]
```diff
@@ -354,8 +354,18 @@ copyDependencyToDir <- function(dependency, outputDir, mustWork = TRUE) {
...
-    unlist(dependency[c('script', 'stylesheet', 'attachment')])
+  if (dependency$all_files)
+    files <- list.files(dir)
+  else {
+    file_list <- c(dependency[['script']],
+                   dependency[['stylesheet']],
+                   dependency[['attachment']])
+    if (anyUnnamed(file_list))
+      files <- vapply(file_list,
+                      function(f) if (is.list(f)) f[['src']] else f,
```
cpsievert, line 365: Does this correctly handle the case of `htmlDependency("foo", "1.0", src="", script=list(list(src="file.js")))`? I think the tests should verify that the files were actually copied in order to verify this

R/html_dependency.R
```diff
@@ -354,8 +354,18 @@ copyDependencyToDir <- function(dependency, outputDir, mustWork = TRUE) {
...
 
-  files <- if (dependency$all_files) list.files(dir) else {
-    unlist(dependency[c('script', 'stylesheet', 'attachment')])
+  if (dependency$all_files)
+    files <- list.files(dir)
+  else {
+    file_list <- c(dependency[['script']],
+                   dependency[['stylesheet']],
+                   dependency[['attachment']])
+    if (anyUnnamed(file_list))
```
cpsievert, line 363: I think it'd be clearer what's going on here if we were to have a helper function "pluck" the files from each slot first, then put them together. Something like 

```r
pluck_src <- function(x) {
  if (is.character(x)) return(x)
  if (is.list(x) && "src" %in% names(x)) return(x$src) 
  if (is.list(x)) return(pluck_src(x))
  NULL
}

files <- c(
  pluck_src(dep$script),
  pluck_src(dep$stylesheet),
  pluck_src(dep$attachment),
)
```

dmurdoch: The function makes sense.   I'll add that.

R/html_dependency.R
dmurdoch: Yes, that's worth testing.  I'll add that.

dmurdoch:
Just committed changes as per suggestions.

dmurdoch:
Just a note in case there are more comments:  I'm going offline soon, and won't be available for a week or two.  I'll address other comments when I get back.  If they're urgently needed before that, someone else will have to take over.

R/utils.R
```diff
@@ -151,3 +151,12 @@ anyUnnamed <- function(x) {
   # List with name attribute; check for any ""
   any(!nzchar(nms))
 }
+
+# Get the source filename out of a script or similar
+# entry.
+pluck_src <- function(x) {
+  if (is.character(x)) return(x)
+  if (is.list(x) && "src" %in% names(x)) return(x$src)
+  if (is.list(x)) return(vapply(x, pluck_src, ""))
```
wch, line 160: I'm finding it a bit challenging to understand what exactly this function is trying to do. I think it's in part because the shape of the data going in is pretty loose -- that's something that has long been a cause of issues when dealing with html dependency objects.

In this particular case, I'm wondering about the `NULL` return and the `vapply()`. If this `vapply()` is called on the object and one of the items returns `NULL`, that will result in an error. Will it ever get to the `NULL` at the end of the function? If not, it would make sense to have it throw an error with a helpful message.

R/html_dependency.R
```diff
@@ -366,11 +369,11 @@ copyDependencyToDir <- function(dependency, outputDir, mustWork = TRUE) {
   }
 
   srcfiles <- file.path(dir, files)
-  if (any(!file.exists(srcfiles))) {
+  if (any(bad <- !file.exists(srcfiles))) {
```
wch, line 372: I think the assignment in an `if()` can be pretty confusing, so this should be moved out. (I'll take care of this.)

R/html_dependency.R
```diff
@@ -366,11 +369,11 @@ copyDependencyToDir <- function(dependency, outputDir, mustWork = TRUE) {
...
     stop(
       sprintf(
         "Can't copy dependency files that don't exist: '%s'",
```
wch, line 375: I think the assignment in an `if()` can be pretty confusing, so this should be moved out.

GitHub wouldn't let me add this suggestion in one block (something about not allowing suggestions with deleted lines) so it's in two parts.

```suggestion
  missing_srcfiles <- !file.exists(srcfiles)
  if (any(missing_srcfiles)) {
    stop(
      sprintf(
        "Can't copy dependency files that don't exist: '%s'",
```

R/html_dependency.R
```diff
@@ -366,11 +369,11 @@ copyDependencyToDir <- function(dependency, outputDir, mustWork = TRUE) {
...
-        paste(srcfiles, collapse = "', '")
+        paste(srcfiles[bad], collapse = "', '")
```
wch, line 376:
```suggestion
        paste(srcfiles[missing_srcfiles], collapse = "', '")
```

wch:
I made some changes to the `pluck_src` function and merged into main manually.

wch:
This would much simpler and easier to implement if the `htmlDependency` function normalized the shape of the data structures on the way in. We do it here in TypeScript: https://github.com/rstudio/shiny/blob/474f1400/srcts/src/shiny/render.ts#L355-L467. Maybe someday. Unfortunately, there might code out there that will break because it assumes the older format (just a character vector, not a list like `script = list(src="abc"))`)

PR #342, opened 2022-10-21
Change `fragment.only = TRUE` to `options = '-standalone'`

yihui:
This change will be required in a future version of **markdown**.

I'll let you know when it's ready to merge. Thanks!

### DIFF
--- a/R/tags.R
+++ b/R/tags.R
@@ -1757,7 +1757,12 @@ includeText <- function(path) {
 #' @rdname include
 #' @export
 includeMarkdown <- function(path) {
-  html <- markdown::markdownToHTML(path, options = 'fragment_only')
+  # markdown > v1.2 has removed the fragment.only argument
+  html <- if ("fragment.only" %in% names(formals(markdown::markdownToHTML))) {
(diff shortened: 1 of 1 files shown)


R/tags.R
```diff
@@ -1757,7 +1757,7 @@ includeText <- function(path) {
 #' @rdname include
 #' @export
 includeMarkdown <- function(path) {
-  html <- markdown::markdownToHTML(path, fragment.only=TRUE)
+  html <- markdown::markdownToHTML(path, options = 'fragment_only')
```
cpsievert, line 1760: I think we'll probably want to do something like  

```r
html <- if (is_available("markdown", "TBD") {
  markdown::markdownToHTML(path, options = 'fragment_only')
} else {
  markdown::markdownToHTML(path, fragment.only=TRUE)
}
```

Since `{markdown}` is only a Suggests?

yihui: Good idea.

yihui:
This is ready to merge now. Thanks!

cpsievert:
Thank you!

PR #351, opened 2022-11-01
Implement `tagQuery()$matches(fn)`

### LINKED ISSUE #350: Add `tagQuery()` `$matches(cssSelector)` functionality
After another discussion, `tagQuery()` will add a method like `$matches()` similar to a vectorized form of [`Element.matches(cssSelector)`](https://developer.mozilla.org/en-US/docs/Web/API/Element/matches).

No alterations will be made to tagAppendAttributes(). Users can use `$matches()` to help determine their desired behavior in two steps.

_Originally posted by @schloerke in https://github.com/rstudio/htmltools/issues/334#issuecomment-1298890730_
      

-----------------------

Should there also be the `tag**` helper method? `tagMatches(x, cssSelector)`?

schloerke:
Fixes #350 

```r
(html <- tagList(div(), span()))
#> <div></div>
#> <span></span>
tagQ <- tagQuery(html)
tagQ$matches("span")
#> [1] FALSE TRUE
tagQ$matches(function(el, i) { el$name == "span" })
#> [1] FALSE TRUE
```

### DIFF
--- a/R/tag_query.R
+++ b/R/tag_query.R
@@ -631,14 +631,21 @@ tagQuery_ <- function(
             tagQueryFindClosest(selected_, cssSelector)
           )
         },
-        #' ### Custom filter
+        #' ### Filter
         #'
+        #' * `$matches(fn)`: For each for the selected tags, return `TRUE` if
+        #'  `fn(el)` returns `TRUE`. In addition to an R function with two
+        #'   arguments (the selected tag `x` and the index `i`), `fn` may also
+        #'   be a valid CSS selector.
+        matches = function(fn) {
+          tagQueryMatches(selected_, fn)
+        },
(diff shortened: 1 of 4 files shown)


R/tag_query.R
```diff
@@ -631,14 +631,21 @@ tagQuery_ <- function(
...
           )
         },
-        #' ### Custom filter
+        #' ### Filter
         #'
+        #' * `$matches(fn)`: For each of the selected tags, return `TRUE` if
+        #'  `fn(el)` returns `TRUE`. In addition to an R function with two
+        #'   arguments (the selected tag `x` and the index `i`), `fn` may also
+        #'   be a valid CSS selector.
+        matches = function(fn) {
```
cpsievert, line 640: I feels a bit weird to have the 1st method under a "Filter" section be `matches` and the 2nd one be `filter`

PR #367, opened 2023-02-21
Avoid testing messages generated in rlang

lionel-:
The error message in case of empty arguments has changed slightly in the next version of rlang that is about to be sent to CRAN. This causes some tests in htmltools to fail because they check the exact content of error messages generated in a third party package. To avoid this, I changed the expectations to error snapshots.

### DIFF
--- a/tests/testthat/_snaps/tags.md
+++ b/tests/testthat/_snaps/tags.md
@@ -1,3 +1,18 @@
+# Hanging commas don't break things
+
+    Code
+      (expect_error(as.character(div("one", , ))))
+    Output
+      <error/rlang_error>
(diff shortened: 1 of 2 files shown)


cpsievert:
Thanks @lionel-. I think we'd prefer to still run these still on CRAN (and in a way that isn't restricting the rlang version), so I'll close this in preference of #368

lionel-:
Note that the `expect_error()` inside the snapshot does run on CRAN, it's just the contents of the messages that are not checked on CRAN. I agree that the dependency on the rlang version is not ideal, I suggested an alternative in #368.

PR #368, opened 2023-02-21
Update tests to accomodate new trailing comma error message from rlang

cpsievert:
Supersedes #367.

This change makes the tests pass with both the old rlang <1.0.7 error message:

```r
> rlang::list2(,1)
! Argument 1 is empty
```

As well as the new rlang >= 1.0.7

```r
> rlang::list2(,1)
! Argument 1 can't be empty.
```

### DIFF
--- a/tests/testthat/test-tags.r
+++ b/tests/testthat/test-tags.r
@@ -26,9 +26,16 @@ test_that("Hanging commas don't break things", {
   expect_equal(as.character(tagList("hi",)), "hi")
   expect_equal(as.character(div("one",)), "<div>one</div>")
   # Multiple commas still throw
-  expect_error(as.character(div("one",,)), "empty")
+  err_comma_multiple <- expect_error(as.character(div("one",,)))
   # Non-trailing commas still throw
-  expect_error(as.character(div(,"one",)), "empty")
+  err_comma_leading <- expect_error(as.character(div(,"one",)))
+
+  # rlang > 1.0.6 changed the error message, so only run
+  # snapshot testing of the error message with the new version
+  skip_if_not_installed("rlang", "1.0.6.9000")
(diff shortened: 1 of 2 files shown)


tests/testthat/test-tags.r
```diff
@@ -26,9 +26,9 @@ test_that("Hanging commas don't break things", {
   expect_equal(as.character(tagList("hi",)), "hi")
   expect_equal(as.character(div("one",)), "<div>one</div>")
   # Multiple commas still throw
-  expect_error(as.character(div("one",,)), "is empty")
+  expect_error(as.character(div("one",,)), "empty")
```
lionel-, line 29: I'd prefer if you didn't check the content of messages that comes from rlang (a third party package). Would you consider to at least add a `skip_on_cran()` so that changes won't cause CRAN failures? (I agree that it's unlikely the message will be changed in a way that doesn't involve the word "empty" but this should be a basic hygiene rule of package checks interop to ensure smooth releases)

To avoid the dependency on the rlang version and a hard-check on messages that you don't own, and still check on CRAN that an error occurs and still monitor the contents of the message, I think this approach would work:

- A message-less `expect_error()` first.
- A `skip_if_not_installed()` with minimal rlang version.
- Then error snapshots under that trailing section. The snapshot gives you more information, for example that the error call that shows up is not very helpful to the user. This made me realise that `dots_list()` should have `error_call` and `error_arg` arguments.

The last two steps can be replaced by a skip-on-cran and an error message expectation if you prefer.

cpsievert: Ahh, thanks for the suggestion (and pointing out my misunderstanding)!

lionel-: Thanks!

PR #414, opened 2023-10-16
Improve performance of tags functions

mgirlich:
Functions like `tags$tr()` can easily be a little faster.

```r
library(htmltools)

set.seed(1)
n <- 50e3
data <- vroom::gen_character(n)

f <- function(data) {
  for (i in seq_along(data)) {
    tags$tr(data[[i]])
  }
}

bench::mark(f(data))

# before
# # A tibble: 1 × 13
#   expression      min   median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc total_time
#   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl> <int> <dbl>   <bch:tm>
# 1 f(data)       1.37s    1.37s     0.730    41.2KB     27.8     1    38      1.37s

# after
# # A tibble: 1 × 13
#   expression      min   median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc total_time
#   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl> <int> <dbl>   <bch:tm>
# 1 f(data)       830ms    830ms      1.20    30.1KB     30.1     1    25      830ms
```

### DIFF
--- a/R/tags.R
+++ b/R/tags.R
@@ -170,7 +170,12 @@ nullOrEmpty <- function(x) {
 
 # Given a vector or list, drop all the NULL or length-0 items in it
 dropNullsOrEmpty <- function(x) {
-  x[!vapply(x, nullOrEmpty, FUN.VALUE=logical(1))]
+  ns <- lengths(x) == 0
+  if (any(ns)) {
+    x <- x[!ns]
+  }
+
+  x
 }
 
 isResolvedTag <- function(x) {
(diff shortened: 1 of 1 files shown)


R/tags.R
```diff
@@ -170,7 +170,12 @@ nullOrEmpty <- function(x) {
 
 # Given a vector or list, drop all the NULL or length-0 items in it
 dropNullsOrEmpty <- function(x) {
-  x[!vapply(x, nullOrEmpty, FUN.VALUE=logical(1))]
+  ns <- lengths(x) == 0
```
cpsievert, line 173: From a quick look, it seems `lengths()` was added [in R 3.2.0](https://www.rdocumentation.org/packages/base/versions/3.2.0), so we'll want to put `Depends: R (>= 3.2.0)` in the DESCRIPTION

Also, this change would mean `nullOrEmpty` no longer needs to exist (so please remove it)

PR #416, opened 2023-10-17
Implement `tag()` in C

mgirlich:
Supersedes #414.

I think it is worth implementing `tag()` in C to make it fast. Especially when using `tag()` directly instead of `tags$...` this is now quite fast.

```r
library(htmltools)

set.seed(1)
n <- 50e3
data <- vroom::gen_character(n)

f_tags <- function(data) {
  for (i in seq_along(data)) {
    tags$tr(data[[i]])
  }
}
f_tag <- function(data) {
  for (i in seq_along(data)) {
    tag("tr", list(data[[i]]))
  }
}

bench::mark(
  tags = f_tags(data),
  tag = f_tag(data),
)

# before
# # A tibble: 2 × 13
#   expression      min   median `itr/sec` mem_alloc `gc/sec`
#   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
# 1 tags          1.24s    1.24s     0.810    41.2KB     13.0
# 2 tag        637.34ms 637.34ms     1.57     15.2KB     14.1

# after
# # A tibble: 2 × 13
#   expression      min  median `itr/sec` mem_alloc `gc/sec` n_itr
#   <bch:expr> <bch:tm> <bch:t>     <dbl> <bch:byt>    <dbl> <int>
# 1 tags        426.8ms   441ms      2.27    15.7KB     29.5     2
# 2 tag          89.8ms    91ms     10.8     15.2KB     43.3     6
```


src/tag.c
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  UNPROTECT(2);
+  return out;
+}
+
+SEXP new_tag(SEXP tagName, SEXP varArgs, SEXP noWS, SEXP renderHook) {
+  R_xlen_t n = Rf_xlength(varArgs);
+
+  // TODO validate that varArgs is a list
```
mgirlich, line 27: It is documented that `varArgs` must be a list but there is a test where this is not the case. Should we allow `varArgs` to be e.g. a character and simply wrap it in a list?

tests/testthat/test-tags.r
```diff
@@ -386,6 +386,7 @@ test_that("Old tags predating rlang::list2 can still be rendered", {
 })
 
 test_that("tag with noWS works",{
+  skip("should tag accept only lists?")
```
mgirlich, line 389: see above

PR #421, opened 2023-11-09
`bindFillRole()` should attach its `HTMLDependency()` to fill items

cpsievert:
With this change, something like this now works:

```r
div(class = "html-fill-container", bindFillRole(div(), item = TRUE))
```

The importance of which has increased since the emergence of `div(bslib::as_fillable_container())`, which only adds the HTML classes (not the dependencies).

And since `bslib::page_fillable()` uses `as_fillable_container()`, this currently doesn't work:

```r
library(shiny)
library(plotly)
library(bslib)

ui <- page_fillable(
  plotlyOutput("p")
)

server <- function(input, output) {
  output$p <- renderPlotly({
    plotly::plot_ly()
  })
}

shinyApp(ui, server)
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,10 @@
 # htmltools (development version)
 
+## Bug fixes
+
+* `bindFillRole()` now attaches its `HTMLDependency()` to fill items, thus reducing the possibility of filling layout breaking due to missing CSS. (#421) 
+
+
 # htmltools 0.5.7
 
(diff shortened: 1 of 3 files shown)


gadenbuie (approved):
This looks great to me. I was assuming that `bindFillRole()` always added the fill dependency, not just to containers, so I think making that explicitly the case is a positive move.

PR #422, opened 2023-11-15
Remove dependency on ellipsis.

olivroy:
Replace ellipsis::check_dots_empty() by rlang::check_dots_empty()

Since there is `#' @import rlang`, removing ns suffices, as rlang has check_dots_empty()

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -24,7 +24,6 @@ Depends:
 Imports:
     base64enc,
     digest,
-    ellipsis,
     fastmap (>= 1.1.0),
     grDevices,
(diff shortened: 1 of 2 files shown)


gadenbuie (approved):
Thanks @olivroy!

And a note that `check_dots_empty()` was added in [rlang 1.0.0](https://rlang.r-lib.org/news/index.html#argument-intake-1-0-0). htmltools already depends on `rlang >= 1.0.0` so this change is compatible with htmltools' dependency list.

cpsievert:
Thanks!

PR #448, opened 2026-05-27
Clearer error when include*() helpers are given a missing file

### LINKED ISSUE #1757: Vague Error if CSS File Missing
If the file provided as an argument to `includeCSS` has been moved, but the file path forgotten to be updated in the application, the Shiny application fails to start and the error message "Error: cannot open the connection" is displayed. Could the error reporting be improved for this and similar scenarios? Perhaps simply using `file.exists` in the `includeCSS`, `includeScipt`, `includeHTML` functions would help to improve the UX.

samuelbharti:
Fixes rstudio/shiny#1757

## Summary

- `includeCSS()`, `includeHTML()`, `includeText()`, `includeMarkdown()`, and `includeScript()` previously surfaced the generic `cannot open the connection` message from `readLines()` when given a non-existent `path`.
- This PR adds an up-front `file.exists()` check via a new internal helper `check_include_path()`, which raises a structured `rlang::abort()` naming the missing file.
- `NEWS.md` updated.

## Test plan

- [ ] `devtools::test()` passes
- [ ] Manually verify each helper raises the new error when given a missing path
- [ ] Manually verify each helper still works for an existing file

### DIFF
--- a/R/tags.R
+++ b/R/tags.R
@@ -1733,6 +1733,7 @@ knit_print.html_dependency <- knit_print.shiny.tag
 #' @aliases includeHTML
 #' @export
 includeHTML <- function(path) {
+  check_include_path(path, "HTML file")
   lines <- readLines(path, warn=FALSE, encoding='UTF-8')
 
   if (detect_html_document(lines)) {
@@ -1781,6 +1782,7 @@ detect_html_document <- function(lines) {
 #' @rdname include
 #' @export
 includeText <- function(path) {
+  check_include_path(path, "Text file")
   lines <- readLines(path, warn=FALSE, encoding='UTF-8')
   return(paste8(lines, collapse='\n'))
 }
@@ -1790,6 +1792,7 @@ includeText <- function(path) {
 #' @rdname include
 #' @export
 includeMarkdown <- function(path) {
+  check_include_path(path, "Markdown file")
   # markdown >= v1.3 switched from markdownToHTML() to mark()
   html <- if (packageVersion("markdown") < "1.3") {
     markdown::markdownToHTML(path, fragment.only = TRUE)
@@ -1804,12 +1807,7 @@ includeMarkdown <- function(path) {
 #' @rdname include
 #' @export
 includeCSS <- function(path, ...) {
-  if (!file.exists(path)) {
-    rlang::abort(c(
-      "CSS file does not exist.",
-      "x" = paste0("Path: ", path)
-    ))
-  }
+  check_include_path(path, "CSS file")
   lines <- readLines(path, warn=FALSE, encoding='UTF-8')
   args <- dots_list(...)
   if (is.null(args$type))
@@ -1821,8 +1819,18 @@ includeCSS <- function(path, ...) {
 #' @rdname include
 #' @export
 includeScript <- function(path, ...) {
+  check_include_path(path, "Script file")
   lines <- readLines(path, warn=FALSE, encoding='UTF-8')
   return(tags$script(HTML(paste8(lines, collapse='\n')), ...))
+}
+
+check_include_path <- function(path, what = "File") {
+  if (!file.exists(path)) {
+    rlang::abort(c(
+      paste0(what, " does not exist."),
+      "x" = paste0("Path: ", path)
+    ))
+  }
 }
 
 #' Include content only once
--- a/R/tags.R
+++ b/R/tags.R
@@ -1804,6 +1804,12 @@ includeMarkdown <- function(path) {
 #' @rdname include
 #' @export
 includeCSS <- function(path, ...) {
+  if (!file.exists(path)) {
+    rlang::abort(c(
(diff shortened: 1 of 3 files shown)


R/tags.R
```diff
@@ -1815,10 +1819,20 @@ includeCSS <- function(path, ...) {
...
 }
 
+check_include_path <- function(path, what = "File") {
+  if (!file.exists(path)) {
+    rlang::abort(c(
+      paste0(what, " does not exist."),
+      "x" = paste0("Path: ", path)
+    ))
+  }
+}
```
schloerke: Consider passing `call = caller_env()` so the error attributes to the public helper rather than to `check_include_path()`.

Without it, the error header reads `` Error in `check_include_path()`: `` instead of `` Error in `includeCSS()`: ``, which makes the source of the problem harder for users to find.

```suggestion
check_include_path <- function(path, what = "File", call = rlang::caller_env()) {
  if (!file.exists(path)) {
    rlang::abort(
      c(paste0(what, " does not exist."), "x" = paste0("Path: ", path)),
      call = call
    )
  }
}
```

schloerke: Validating only `file.exists()` leaves two edge cases that still produce confusing errors:

- `file.exists(NULL)` returns `logical(0)`, so `if (!file.exists(path))` itself errors with "argument is of length zero" — we've swapped one unhelpful message for another.
- `file.exists()` returns `TRUE` for directories, so `readLines()` will still throw its cryptic error if `path` is a directory.

A small guard handles both:

```r
check_include_path <- function(path, what = "File", call = rlang::caller_env()) {
  if (!rlang::is_string(path) || !nzchar(path)) {
    rlang::abort(
      paste0(what, " path must be a single non-empty string."),
      call = call
    )
  }
  if (!file.exists(path) || dir.exists(path)) {
    rlang::abort(
      c(paste0(what, " does not exist."), "x" = paste0("Path: ", path)),
      call = call
    )
  }
}
```

tests/testthat/test-tags.r
```diff
@@ -1226,3 +1226,20 @@ test_that("includeHTML() warns if full document is detected", {
...
+  expect_error(includeText(missing),     "Text file does not exist")
+  expect_error(includeScript(missing),   "Script file does not exist")
+  skip_if_not_installed("markdown")
+  expect_error(includeMarkdown(missing), "Markdown file does not exist")
+})
+
+test_that("include*() error message names the missing path", {
+  missing <- tempfile("does-not-exist-", fileext = ".txt")
+  expect_error(includeCSS(missing), missing, fixed = TRUE)
+})
```
schloerke, line 1264: Two small test cleanups:

1. `skip_if_not_installed("markdown")` currently sits in the middle of the test, after the CSS/HTML/Text/Script expectations have already run. Convention is to put `skip_*` at the top of a test. Since the skip only gates the markdown expectation, it would be cleaner to lift that into its own `test_that()`.
2. The second test (`include*() error message names the missing path`) only re-checks `includeCSS` and exercises the same code path as the test above — it can be removed.

```suggestion
test_that("include*() helpers raise a clear error when the file is missing", {
  missing <- tempfile("does-not-exist-", fileext = ".txt")
  expect_false(file.exists(missing))

  expect_error(includeCSS(missing),    "CSS file does not exist")
  expect_error(includeHTML(missing),   "HTML file does not exist")
  expect_error(includeText(missing),   "Text file does not exist")
  expect_error(includeScript(missing), "Script file does not exist")
})

test_that("includeMarkdown() raises a clear error when the file is missing", {
  skip_if_not_installed("markdown")
  missing <- tempfile("does-not-exist-", fileext = ".txt")
  expect_error(includeMarkdown(missing), "Markdown file does not exist")
})
```

R/tags.R
samuelbharti: Done in daa8452 — `check_include_path()` now takes `call = rlang::caller_env()` and forwards it to `rlang::abort()`, so the error header now reads `` Error in `includeCSS()`: `` (or whichever helper).

Added a test that catches the condition and asserts the deparsed `call` contains the helper name.

samuelbharti: Done in 3c635cf — adopted the suggested guard. `check_include_path()` now:

- Errors with "path must be a single non-empty string" when `path` is not a `rlang::is_string()` or is `""`. This catches `NULL`, `character(0)`, multi-element vectors, `""`, and `NA_character_`.
- Treats a directory at `path` the same as a missing file (raises the existing "does not exist" error) so `readLines()` never sees a directory.

Added tests for each of those edge cases.

tests/testthat/test-tags.r
samuelbharti: Done in 9de3434 — `includeMarkdown()` now lives in its own `test_that()` with `skip_if_not_installed("markdown")` at the top, and the redundant second test has been removed.

schloerke (approved):

